PR #10, opened 2019-06-21
Integration test

red1408:
Uses dcm4che-store-scu as client to upload example dcm via import adapter then downloads it back via curl and compares to original.

### DIFF
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -26,7 +26,7 @@
 # run adapter, close when connection established on _CLOSE_ADAPTER_PORT
 - name: 'gradle:5.4.1-jdk8'
   id: gradle-run-adapter
-  args: [ 'bash', '../integration_test/run_adapter_and_wait.sh', '${_STAGE}', '${_PROJECT}', '${_LOCATION}', '${_DATASET}', '${_STORE_NAME}', '${_ADAPTER_PORT}', '${_CLOSE_ADAPTER_PORT}']
+  args: [ 'bash', '../integration_test/run_adapter_and_wait.sh', '${_VERSION}', '${_PROJECT}', '${_LOCATION}', '${_DATASET}', '${_STORE_NAME}', '${_ADAPTER_PORT}', '${_CLOSE_ADAPTER_PORT}']
   dir: import
   waitFor: ['-']
 
@@ -97,7 +97,7 @@
   args:
   - bash
   - integration_test/curl-dcm.sh
-  - "https://healthcare.googleapis.com/${_STAGE}/projects/${_PROJECT}/locations/${_LOCATION}/datasets/${_DATASET}/dicomStores/${_STORE_NAME}/dicomWeb/studies/1.3.6.1.4.1.5962.1.167.1174395400.11033.0/series/1.3.6.1.4.1.5962.1.176.1174395403.11033.0/instances/1.3.6.1.4.1.5962.1.178.1174395404.11033.0"
+  - "https://healthcare.googleapis.com/${_VERSION}/projects/${_PROJECT}/locations/${_LOCATION}/datasets/${_DATASET}/dicomStores/${_STORE_NAME}/dicomWeb/studies/1.3.6.1.4.1.5962.1.167.1174395400.11033.0/series/1.3.6.1.4.1.5962.1.176.1174395403.11033.0/instances/1.3.6.1.4.1.5962.1.178.1174395404.11033.0"
   - integration_test/downloaded.dcm
   waitFor:
   - run-store-scu
@@ -127,7 +127,7 @@
   - diff-dcm
 
 substitutions:
-  _STAGE: v1beta1
+  _VERSION: v1beta1
   _PROJECT: healthcare-api-215503
   _LOCATION: europe-west2
   _DATASET: example-west
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -77,7 +77,7 @@
   - mvn
   - 'exec:java'
   - '-Dexec.mainClass=org.dcm4che3.tool.storescu.StoreSCU'
-  - '-Dexec.args=-c IMPORTADAPTER@step_3:${_ADAPTER_PORT} ../../../integration_test/example.dcm'
+  - '-Dexec.args=-c IMPORTADAPTER@${_ADAPTER_RUN_STEP}:${_ADAPTER_PORT} ../../../integration_test/example.dcm'
   dir: "dcm4che/dcm4che-tool/dcm4che-tool-storescu"
   waitFor:
   - wait-for-adapter
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -26,7 +26,7 @@
 # run adapter, close when connection established on _CLOSE_ADAPTER_PORT
 - name: 'gradle:5.4.1-jdk8'
   id: gradle-run-adapter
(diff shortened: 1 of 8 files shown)


integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - delete
+  - ${_STORE_NAME}
+  - "--dataset=${_DATASET}"
+  - "--location=${_LOCATION}"  
+  - "--quiet"
+  waitFor:
+  - diff-dcm
+
+substitutions:
+  _STAGE: v1beta1
```
danielbeaudreau: Can we call this _VERSION instead of _STAGE

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - create
+  - ${_STORE_NAME}
+  - "--dataset=${_DATASET}"
+  - "--location=${_LOCATION}"  
+  - "--quiet"
+  waitFor: 
+  - cleanup-dicom-store
+
+
+- name: 'busybox:latest'
```
danielbeaudreau: Just wondering, why do we need busybox here?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+- name: "maven:3.6.0-jdk-8"
+  id: run-store-scu
+  volumes:
+  - name: 'vol1'
+    path: '/root/.m2'
+  args:
+  - mvn
+  - 'exec:java'
+  - '-Dexec.mainClass=org.dcm4che3.tool.storescu.StoreSCU'
+  - '-Dexec.args=-c IMPORTADAPTER@step_3:${_ADAPTER_PORT} ../../../integration_test/example.dcm'
```
danielbeaudreau: Can you use the variable ${_ADAPTER_RUN_STEP} instead of step_3?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - "--quiet"
+  waitFor: 
+  - cleanup-dicom-store
+
+
+- name: 'busybox:latest'
+  id: wait-for-adapter
+  args:
+  - ash
+  - integration_test/wait_for_port.sh
```
danielbeaudreau: Instead of having the wait-for-adapter step, since run_adapter_and_wait.sh already waits for the adapter to start, do we need wait_for_port.sh or could we delete this step and add a waitFor in run-store-scu for build-store-scu, create-dicom-store, and gradle-run-adapter? You might have to modify run_adapter_and_wait.sh to wait for both the adapter port and the close port

integration_test/curl-dcm.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# url, output_file
+curl -X GET \
+     -H "Authorization: Bearer "$(gcloud auth print-access-token) \
+     $1 \
+     |tail -c+137 | head -c-68 >$2
```
danielbeaudreau: Are you using head and tail here to strip the multipart boundaries? You can instead specify the accept header application/dicom; transfer-syntax=* to get the instance without a boundary.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - diff-dcm
+
+substitutions:
+  _STAGE: v1beta1
+  _PROJECT: healthcare-api-215503
+  _LOCATION: europe-west2
+  _DATASET: example-west
+  _STORE_NAME: IntegrationTestStore
+  _ADAPTER_PORT: '2575'
+  _CLOSE_ADAPTER_PORT: '3000'
```
danielbeaudreau: I wasn't aware that the adapter shuts down when a packet is sent to port 3000. Is this functionality from the dcm4che library it's based on?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -0,0 +1,137 @@
...
+  _ADAPTER_RUN_STEP: 'step_3'
```
danielbeaudreau: Just wondering, why is the hostname step_3 instead of localhost?

integration_test/cloudbuild-integration-test.yaml
red1408: Ok, changed to _VERSION

integration_test/cloudbuild-integration-test.yaml
red1408: It's a very small image, so it just seems to make sense for utility steps. But ultimately any image with nc (netcat) would do.

integration_test/cloudbuild-integration-test.yaml
red1408: We can't use cloud build wait feature for gradle-run-adapter because I'm waiting for adapter to be up and running, not for it's step is finish.
So the solution I came up with is to run adapter, wait for it's dimse port to be up, run c-store scu, then use connection on close_port as signal to finish adapter cloud build step.

integration_test/curl-dcm.sh
red1408: Yeah, tail and part solution is ugly, changed to header. Weird though that stow-rs standard says that response is always multipart, even for single instance.

integration_test/cloudbuild-integration-test.yaml
red1408: No it isn't. It's from run_adapter_and_wait.sh script. I just run adapter in background (&) and wait on port 3000 via nc to gracefully finish this cloud build step. 
Adapter itself has no shutdown feature, and adding one just for integration test didn't seem right either.

integration_test/cloudbuild-integration-test.yaml
red1408: step_0, step_1 etc is how cloud build names docker containers that run each step, these are also their dns names on docker network. There doesn't seem a way to redefine these names in cloud build.

integration_test/cloudbuild-integration-test.yaml
red1408: Yes, just missed this one. Will fix.

red1408:
Trigger failed because integration test is in sub-folder ( default trigger looks for cloudbuild.yaml in root). My original intention was to add default cloudbuild.yaml with simple 'gradle build' instruction as unit test runner, and have a separate trigger for integration test.
Or I could rename and move integration test yaml to root.
Which way would be preferable?

danielbeaudreau:
If you can add a separate trigger for the integration test, that would be preferable I think. The default unit test runner looks good.

red1408:
I don't have rights to create triggers from github.
Screenshot from 2019-06-26 13-46-33

integration_test/setup_dataset_and_store.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+  gcloud alpha healthcare datasets create $2 "--location="$1 "--quiet" || exit $?
+fi
+
+gcloud alpha healthcare dicom-stores describe $3 "--dataset="$2 "--location="$1 "--quiet" >/dev/null
+if [ $? -eq 0 ];
+then
+  gcloud alpha healthcare dicom-stores delete $3 "--dataset="$2 "--location="$1 "--quiet" || exit $?
+fi
+gcloud alpha healthcare dicom-stores create $3 "--dataset="$2 "--location="$1 "--quiet"
```
danielbeaudreau: Should this have "|| exit $?" at the end too?

danielbeaudreau:
I have the permissions to create new triggers. Please merge this PR (after taking a look at the above small comment) and then I will then set up a new trigger for this yaml file tomorrow.

integration_test/setup_dataset_and_store.sh
red1408: Last command's exit code is already whole script's exit code, so there is no need to.

danielbeaudreau: Ah, got it.

red1408:
"Only those with write access to this repository can merge pull requests.", which I apparently do not have, so I can't merge.

danielbeaudreau:
Ok, I merged it.

PR #14, opened 2019-07-04
2nd cfind

red1408:
2nd part of adapter improvements:
- cfind
- cfind service test
- attributes util test (used within cfind)

Waiting for 1st part to be merged. Since it contains all commits of 1st part, pull request will show too many changes before 1st one is merged. Most actual changes are in this commit: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/pull/14/commits/a1c1c651c3cfe2da2db84e3448de4316781bc9d9

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@ two components, namely import and export adapter.
 
 ## Import Adapter
 
-The Import Adapter converts incoming DIMSE (C-STORE) requests to DICOMWeb (STOW-RS) and sends them to peer.
+The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests:
+- C-STORE to STOW-RS
+- C-FIND to QIDO-RS (multi-modality C-FIND queries result in 1 QIDO-RS query per modality) and passes converted results back to DIMSE client
 
 For the list of command line flags, see [here](import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java)
 
--- a/README.md
+++ b/README.md
@@ -6,9 +6,11 @@ two components, namely import and export adapter.
 
 ## Import Adapter
 
-The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests:
+The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests and passes the converted results back to the DIMSE client. The following requests are supported:
 - C-STORE to STOW-RS
-- C-FIND to QIDO-RS (multi-modality C-FIND queries result in 1 QIDO-RS query per modality) and passes converted results back to DIMSE client
+- C-FIND to QIDO-RS
+
+Note that any C-FIND query on the ModalitiesInStudy tag will result in 1 QIDO-RS query per modality.
 
 For the list of command line flags, see [here](import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java)
 
--- a/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
+++ b/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
@@ -105,11 +105,11 @@ public class CFindService extends BasicCFindSCP {
           }
           log.info("CFind QidoPath: " + qidoPath);
           try {
-            MonitoringService.addEvent(Event.CFIND_SUB_QIDORS_REQUEST);
+            MonitoringService.addEvent(Event.CFIND_QIDORS_REQUEST);
             JSONArray qidoResult = dicomWebClient.qidoRs(qidoPath);
             qidoResults.add(qidoResult);
           } catch (IDicomWebClient.DicomWebException e) {
-            MonitoringService.addEvent(Event.CFIND_SUB_QIDORS_ERROR);
+            MonitoringService.addEvent(Event.CFIND_QIDORS_ERROR);
             if (e.getCause() instanceof HttpResponseException &&
                 ((HttpResponseException) e.getCause()).getStatusCode() ==
                     HttpStatusCodes.STATUS_CODE_SERVICE_UNAVAILABLE) {
--- a/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
+++ b/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
@@ -1,0 +1,153 @@
+package com.google.cloud.healthcare.imaging.dicomadapter;
+
+import com.google.api.client.http.HttpResponseException;
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.cloud.healthcare.IDicomWebClient;
+import com.google.cloud.healthcare.imaging.dicomadapter.monitoring.Event;
+import com.google.cloud.healthcare.imaging.dicomadapter.monitoring.MonitoringService;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import org.dcm4che3.data.Attributes;
+import org.dcm4che3.data.Tag;
+import org.dcm4che3.data.UID;
+import org.dcm4che3.data.VR;
+import org.dcm4che3.net.Association;
+import org.dcm4che3.net.Commands;
+import org.dcm4che3.net.Dimse;
+import org.dcm4che3.net.Status;
+import org.dcm4che3.net.pdu.PresentationContext;
+import org.dcm4che3.net.service.BasicCFindSCP;
+import org.dcm4che3.net.service.DicomServiceException;
+import org.dcm4che3.util.TagUtils;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class CFindService extends BasicCFindSCP {
+
+  private static Logger log = LoggerFactory.getLogger(CFindService.class);
+
+  private IDicomWebClient dicomWebClient;
+
+  CFindService(IDicomWebClient dicomWebClient) {
+    super(UID.StudyRootQueryRetrieveInformationModelFIND);
+    this.dicomWebClient = dicomWebClient;
+  }
+
+  private static HashMap<String, JSONObject> uniqueResults(List<JSONArray> responses) {
+    HashMap<String, JSONObject> uniqueResults = new HashMap<>();
+    for (JSONArray response : responses) {
+      for (Object result : response) {
+        JSONObject resultJson = (JSONObject) result;
+        String key = getResultKey(resultJson);
+        if (!uniqueResults.containsKey(key)) {
+          uniqueResults.put(key, resultJson);
+        }
+      }
+    }
+    return uniqueResults;
+  }
+
+  private static String getResultKey(JSONObject jsonObject) {
+    return AttributesUtil.getTagValueOrNull(jsonObject,
+        TagUtils.toHexString(Tag.StudyInstanceUID)) + "_" +
+        AttributesUtil.getTagValueOrNull(jsonObject,
+            TagUtils.toHexString(Tag.SeriesInstanceUID)) + "_" +
+        AttributesUtil.getTagValueOrNull(jsonObject,
+            TagUtils.toHexString(Tag.SOPInstanceUID));
+  }
+
+  @Override
+  public void onDimseRQ(Association association,
+      PresentationContext presentationContext,
+      Dimse dimse,
+      Attributes request,
+      Attributes keys) throws IOException {
+    if (dimse != Dimse.C_FIND_RQ) {
+      throw new DicomServiceException(Status.UnrecognizedOperation);
+    }
+
+    MonitoringService.addEvent(Event.CFIND_REQUEST);
+
+    CFindTask task = new CFindTask(association, presentationContext, request, keys);
+    association.getApplicationEntity().getDevice().execute(task);
+  }
+
+  private class CFindTask extends DimseTask {
+
+    private final Attributes keys;
+
+    private CFindTask(Association as, PresentationContext pc,
+        Attributes cmd, Attributes keys) {
+      super(as, pc, cmd);
+
+      this.keys = keys;
+    }
+
+    @Override
+    public void run() {
+      try {
+        if (canceled) {
+          throw new CancellationException();
+        }
+        runThread = Thread.currentThread();
+
+        String[] qidoPaths = AttributesUtil.attributesToQidoPathArray(keys);
+        List<JSONArray> qidoResults = new ArrayList<>();
+        for (String qidoPath : qidoPaths) {
+          if (canceled) {
+            throw new CancellationException();
+          }
+          log.info("CFind QidoPath: " + qidoPath);
+          try {
+            MonitoringService.addEvent(Event.CFIND_SUB_QIDORS_REQUEST);
+            JSONArray qidoResult = dicomWebClient.qidoRs(qidoPath);
+            qidoResults.add(qidoResult);
+          } catch (IDicomWebClient.DicomWebException e) {
+            MonitoringService.addEvent(Event.CFIND_SUB_QIDORS_ERROR);
+            if (e.getCause() instanceof HttpResponseException &&
+                ((HttpResponseException) e.getCause()).getStatusCode() ==
+                    HttpStatusCodes.STATUS_CODE_SERVICE_UNAVAILABLE) {
+              MonitoringService.addEvent(Event.CFIND_ERROR);
+              as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.OutOfResources));
+              return;
+            } else {
+              throw e;
+            }
+          }
+        }
+        HashMap<String, JSONObject> uniqueResults = uniqueResults(qidoResults);
+
+        for (JSONObject obj : uniqueResults.values()) {
+          if (canceled) {
+            throw new CancellationException();
+          }
+          Attributes attrs = AttributesUtil.jsonToAttributes(obj);
+          as.writeDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.Pending), attrs);
+        }
+        as.writeDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.Success));
+      } catch (CancellationException e) {
+        log.info("Canceled CFind", e);
+        MonitoringService.addEvent(Event.CFIND_CANCEL);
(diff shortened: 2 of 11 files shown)


danielbeaudreau:
Hi Egor, a couple updates:
1. The PR for stackdriver logging was merged, please rebase this change accordingly.
2. I found out that our service account doesn't have permissions to create datasets. I filed a bug with instructions to fix the test, using a pre-created dataset: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/issues/16 please try that and see if it works. Thanks

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,219 @@
...
+import org.dcm4che3.net.Status;
+import org.dcm4che3.net.service.DicomServiceException;
+import org.dcm4che3.util.TagUtils;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class AttributesUtil {
+
+  public static String getTagValue(JSONObject json, String tag) throws JSONException {
```
danielbeaudreau: This method doesn't seem to be used, can we delete it?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,219 @@
...
+    for (int tag : attrs.tags()) {
+      if (attrs.containsValue(tag)) {
+        nonEmptyKeys.add(tag);
+      }
+    }
+
+    StringBuilder qidoPath = new StringBuilder();
+    if (nonEmptyKeys.contains(Tag.QueryRetrieveLevel)) {
+      switch (attrs.getString(Tag.QueryRetrieveLevel)) {
+        case "PATIENT":
```
danielbeaudreau: Patient level retrieval isn't supported by the CHC api (or QIDO I think), let's delete this case.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,219 @@
...
+          attrs.setDouble(tag, vr, dValues);
+          break;
+        default:
+          throw new DicomServiceException(Status.ProcessingFailure,
+              "Unsupported VR " + vr.toString());
+      }
+    }
+  }
+
+  private static void setAttributeDataFragment(Attributes attrs, int tag, VR vr, JSONObject element)
```
danielbeaudreau: DataFragment isn't part of the standard, must be dcm4che specific. Also, CHC API does not return inline binary, so this code can be deleted.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,219 @@
...
+    return qidoPath.toString();
+  }
+
+  /**
+   * Converts QIDO_RS json reply to dcm4che Attributes
+   * @param jsonDataset
+   */
+  public static Attributes jsonToAttributes(JSONObject jsonDataset) throws DicomServiceException {
+    Attributes attributes = new Attributes();
+    for (String tag : jsonDataset.keySet()) {
```
danielbeaudreau: The structure of PN VR attributes is different than regular strings (it can contain alphabetic, ideographic, and phonetic components).  Please see this example: http://dicom.nema.org/medical/dicom/current/output/html/part18.html#table_F.3.1-1

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
```diff
@@ -0,0 +1,153 @@
...
+
+        String[] qidoPaths = AttributesUtil.attributesToQidoPathArray(keys);
+        List<JSONArray> qidoResults = new ArrayList<>();
+        for (String qidoPath : qidoPaths) {
+          if (canceled) {
+            throw new CancellationException();
+          }
+          log.info("CFind QidoPath: " + qidoPath);
+          try {
+            MonitoringService.addEvent(Event.CFIND_SUB_QIDORS_REQUEST);
```
danielbeaudreau: I'm a bit confused with the name "CFIND_SUB_QIDORS_REQUEST", not sure what SUB means here, can we shorten these to "CFIND_QIDORS_REQUEST", etc?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,219 @@
...
+   */
+  public static String[] attributesToQidoPathArray(Attributes attrs, String... includeFields)
+      throws IOException {
+    if (!attrs.containsValue(Tag.ModalitiesInStudy)
+        || attrs.getStrings(Tag.ModalitiesInStudy).length <= 1) {
+      return new String[]{attributesToQidoPath(attrs, includeFields)};
+    }
+
+    List<String> result = new ArrayList<>();
+    for (String modality : attrs.getStrings(Tag.ModalitiesInStudy)) {
```
danielbeaudreau: Instead of doing this, can't you add them as comma separated instead? http://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.6.1.2.1
See the last example here. That way you only need to ever make one qido request. Please test it against the CHC API to verify it works.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: It it used in next pull request, for cases where not having data should throw exceptions.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: Ok

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: Comma-separated notation within single parameter doesn't seem to be recognized by healthcare api at all (always empty response), as far as I remember dcm4che took first modality instead.

Repeating same parameter is at least recognized syntax, but returns the following: 
"error": {
            "code": 400,
            "message": "generic::invalid_argument: list of UIDs matching unsupported, tag: ModalitiesInStudy",
            "status": "INVALID_ARGUMENT"
        }

And I get same error for StudyInstanceUID despite the fact that it is used in standard example.

I also forwarded you previous e-mail chain I had on this matter with Dusan.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
red1408: Ok, removed.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: DataFragment is probably dcm4che specific, but inline binary is part of standard. But I couldn't get any from healthcare api to check how it looks there (obviously, since you just told healthcare doesn't support it).
Ok, removed.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: Ok, added special handling for PN VR.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: After looking in more detail: there is a copy of this method in export adapter CStoreSender. Removed it to use AttributesUtil one instead, thus also solving problem of unused code.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return result.toArray(new String[]{});
+  }
+
+  /**
+   * Returns corresponding QIDO-RS path
+   * @param attrs dcm4che Attributes to convert
+   * @param includeFields additonal includeFields for QIDO-RS
+   */
+  public static String attributesToQidoPath(Attributes attrs, String... includeFields)
```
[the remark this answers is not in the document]
red1408: Iterating code doesn't try to use comma notation, it sends tag1=[val1,val2,val3] as tag1=val1&tag1=val2&tag1=val3, which is a syntax by api (and server should support UID list matching according to standard, but currently doesn't).

Only UID lists are part of standard, so I could throw error when parameter VR is not UID list, but didn't see much meaning to doing so (since it's easy to handle server-side by ignoring parameter mentions past 1st or throwing error like it does now). In theory api could later choose to support lists for some other parameters too, even if it's not in standard.

Multi-value ModalitiesInStudy was used in clients commonly enough  (despite not being part of standard) to warrant special case handling.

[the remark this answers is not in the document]
red1408: Ok, will add this.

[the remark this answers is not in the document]
red1408: Ok

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
```diff
@@ -0,0 +1,220 @@
...
+    Attributes attrs = AttributesUtil.jsonToAttributes(jsonObj);
+
+    assertThat(attrs).isEqualTo(new Attributes());
+  }
+
+  @Test
+  public void testJsonToAttributes_integerString() throws Exception {
+    // can be parsed as either string or int.
+    JSONObject jsonObj = new JSONObject("{\"" + TagUtils.toHexString(Tag.InstanceNumber)
+        + "\": {\"vr\": \"IS\",\"Value\": [1]}}");
```
danielbeaudreau: Some integer strings can exceed the max int32 (similar with the DS VR), so can we change the code to map these to strings?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
```diff
@@ -0,0 +1,220 @@
...
+    Attributes attrs = AttributesUtil.jsonToAttributes(jsonObj);
+
+    Attributes expected = new Attributes();
+    expected.setString(Tag.PatientName, VR.PN, "Yamada^Tarou=山田^太郎=やまだ^たろう");
+    assertThat(attrs).isEqualTo(expected);
+  }
+
+  @Test
+  public void testJsonToAttributes_double() throws Exception {
+    // is it reliable (double equality comparison)?
```
danielbeaudreau: Given that it runs in the JVM I think double comparison will be repeatable here, so probably ok to remove this comment.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
```diff
@@ -0,0 +1,220 @@
...
+    Attributes attrs = AttributesUtil.jsonToAttributes(jsonObj);
+
+    Attributes expected = new Attributes();
+    expected.setDouble(Tag.EventTimeOffset, VR.FD, 1.25);
+    assertThat(attrs).isEqualTo(expected);
+  }
+
+  @Test
+  public void testJsonToAttributes_float() throws Exception {
+    // is it reliable (float equality comparison)?
```
danielbeaudreau: And this one.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
danielbeaudreau: Fine to leave this as-is.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -0,0 +1,237 @@
...
+    return valueArray.getString(0);
+  }
+
+  /**
+   * Returns array of QIDO-RS paths (more than 1 if multiple ModalitiesInStudy are present)
+   *
+   * @param attrs dcm4che Attributes to convert
+   * @param includeFields additonal includeFields for QIDO-RS
+   */
+  public static String[] attributesToQidoPathArray(Attributes attrs, String... includeFields)
```
danielbeaudreau: I didn't see any callers of this method specify include fields. Or is this from a follow-up PR?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+    // Associate with peer AE.
+    Association association =
+        associate(serverHostname, serverPort,
+            UID.StudyRootQueryRetrieveInformationModelFIND, UID.ExplicitVRLittleEndian);
+
+    Attributes findData = new Attributes();
+    findData.setString(Tag.QueryRetrieveLevel, VR.CS, "IMAGE");
+
+    // Issue CMOVE
```
danielbeaudreau: typo: should be CFIND

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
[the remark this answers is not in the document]
red1408: Ok will throw exception. But I think it's overly restrictive: even if api doesn't currently support this, it returns sane error which leads to same end result (Dimse Processing Failure).

README.md
```diff
@@ -6,7 +6,9 @@ two components, namely import and export adapter.
 
 ## Import Adapter
 
-The Import Adapter converts incoming DIMSE (C-STORE) requests to DICOMWeb (STOW-RS) and sends them to peer.
+The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests:
+- C-STORE to STOW-RS
+- C-FIND to QIDO-RS (multi-modality C-FIND queries result in 1 QIDO-RS query per modality) and passes converted results back to DIMSE client
```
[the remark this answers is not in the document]

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/DimseTask.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public void onCancelRQ(Association as) {
+    log.info(this.getClass().getSimpleName() + " onCancelRQ");
+
+    if (!canceled) {
+      canceled = true;
+      synchronized (this) {
+        if (runThread != null) {
```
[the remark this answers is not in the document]
red1408: Interrupt doesn't kill the thread, it results in Thread.interrupted() returning true. Methods that check it (pretty much all long operations, like sleep) throw InterruptedException. C-find doesn't contain any, so here it actually doesn't do anything. 
In CMoveTask that shares same base class it's important to interrupt currently ongoing c-store (to avoid waiting until full instance transfer is completed on cancel).

Also note catch and finally blocks - relevant error status (worst case - processing failure) and as.removeCancelRQHandler() call are guaranteed to happen.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
red1408: VRs that are both string and int are already parsed as strings. Added VR.DS unit test just in case.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
red1408: Ok

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
red1408: Ok

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: Used in CFindService: line 100 .

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
red1408: Right, will fix.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/DimseTask.java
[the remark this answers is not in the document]
danielbeaudreau: Got it, thanks. Please add a comment to describe this in detail in the CMove PR.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
danielbeaudreau: What I meant was that no caller specifies the optional includeFields parameter, so it's not used code. My preference is to only submit code which is used somewhere, but not a big deal in this case, I will merge this now so we can start with part 3.

red1408: CMove uses attributesToQidoPath with includefields. attributesToQidoPathArray is not called with  includefields, but it makes sense to leave this parameter for symmetry, I think.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/DimseTask.java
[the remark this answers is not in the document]
red1408: Ok

PR #17, opened 2019-07-10
3rd cmove

red1408:
3rd part of adapter improvements:

- cmove
- cmove service test

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,34 +16,8 @@ Available AET destinations for C-MOVE are configured via an AET dictionary json 
 which can be specified either by using the "--aet_dictionary" command line parameter or 
 specifying the "ENV_AETS_JSON" environment variable.
 
-Here is an example JSON dictionary:
-```shell
-[
-	{
-		"name": "WEASIS", 
-		"host": "localhost", 
-		"port": 11113
-	},
-	{
-		"name": "AESKULAP", 
-		"host": "192.168.0.1", 
-		"port": 11114
-	},
-	{
-		"name": "GINKGO_001",
-		"host": "some.domain.name.com",
-		"port": 6666
-	},
-	...
-]
-```
-
-```shell
-kubectl create configmap aet-dictionary --from-file=AETs.json
-```
-
 The following configuration needs to be added to the dicom-adapter.yaml file to use CMOVE. 
-Please see the (link)[Deployment using Kubernetes] section for more information.
+Please see the [Deployment using Kubernetes](#deployment-using-kubernetes) section for more information.
 ```yaml
 env:
 - name: ENV_AETS_JSON
@@ -51,6 +25,29 @@ env:
     configMapKeyRef:
       name: aet-dictionary
       key: AETs.json
+```
+
+Here is an example JSON dictionary:
+```shell
+[
+	{
+		"name": "DEVICE_A", 
+		"host": "localhost", 
+		"port": 11113
+	},
+	{
+		"name": "DEVICE_B", 
+		"host": "192.168.0.1", 
+		"port": 11114
+	},
+	...
+]
+```
+
+And command to create configmap from it:
+
+```shell
+kubectl create configmap aet-dictionary --from-file=AETs.json
 ```
 
 Note that any C-FIND query on the ModalitiesInStudy tag will result in 1 QIDO-RS query per modality.
@@ -78,7 +75,7 @@ For the list of events logged to Stackdriver for the Import Adapter, see [here](
 The monitored resource is configured as k8s_container, with values set from a combination of environment variables configured via Downward API (pod name, pod namespace and container name) and GCP Metadata (project id, cluster name and location). Defaults to the global resource, if k8s_container can't be configured.
 
 The following configuration needs to be added to the dicom-adapter.yaml file to configure the 
-stackdriver monitoring resource. Please see the (link)[Deployment using Kubernetes] section 
+stackdriver monitoring resource. Please see the [Deployment using Kubernetes](#deployment-using-kubernetes) section 
 for more information.
 ```yaml
 env:
--- a/README.md
+++ b/README.md
@@ -9,6 +9,23 @@ two components, namely import and export adapter.
 The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests and passes the converted results back to the DIMSE client. The following requests are supported:
 - C-STORE to STOW-RS
 - C-FIND to QIDO-RS
+- C-MOVE uses QIDO-RS to determine instances to be transfered, then (per instance) executes WADO-RS to obtain instance data stream and passes it to C-STORE (to C-MOVE destination). 
+
+AET resolution for C-MOVE is configured via AET dictionary json file ("--aet_dictionary" command line parameter or "ENV_AETS_JSON" environment variable). Format: JSON array of objects containing name, host and port.
+
+```shell
+kubectl create configmap aet-dictionary --from-file=AETs.json
+```
+
+Relevant part of yaml:
+```yaml
+env:
+- name: ENV_AETS_JSON
+  valueFrom:
+    configMapKeyRef:
+      name: aet-dictionary
+      key: AETs.json
+```
 
 Note that any C-FIND query on the ModalitiesInStudy tag will result in 1 QIDO-RS query per modality.
 
--- a/README.md
+++ b/README.md
@@ -9,15 +9,41 @@ two components, namely import and export adapter.
 The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests and passes the converted results back to the DIMSE client. The following requests are supported:
 - C-STORE to STOW-RS
 - C-FIND to QIDO-RS
-- C-MOVE uses QIDO-RS to determine instances to be transfered, then (per instance) executes WADO-RS to obtain instance data stream and passes it to C-STORE (to C-MOVE destination). 
+- C-MOVE uses QIDO-RS to determine which instances to transfer, then for each instance executes a 
+WADO-RS request to fetch the instance and a C-STORE request to transfer it to the C-MOVE destination 
 
-AET resolution for C-MOVE is configured via AET dictionary json file ("--aet_dictionary" command line parameter or "ENV_AETS_JSON" environment variable). Format: JSON array of objects containing name, host and port.
+Available AET destinations for C-MOVE are configured via an AET dictionary json file, 
+which can be specified either by using the "--aet_dictionary" command line parameter or 
+specifying the "ENV_AETS_JSON" environment variable.
+
+Here is an example JSON dictionary:
+```shell
+[
+	{
+		"name": "WEASIS", 
+		"host": "localhost", 
+		"port": 11113
+	},
+	{
+		"name": "AESKULAP", 
+		"host": "192.168.0.1", 
+		"port": 11114
+	},
+	{
+		"name": "GINKGO_001",
+		"host": "some.domain.name.com",
+		"port": 6666
+	},
+	...
+]
+```
 
 ```shell
 kubectl create configmap aet-dictionary --from-file=AETs.json
 ```
 
-Relevant part of yaml:
+The following configuration needs to be added to the dicom-adapter.yaml file to use CMOVE. 
+Please see the (link)[Deployment using Kubernetes] section for more information.
 ```yaml
 env:
 - name: ENV_AETS_JSON
@@ -51,7 +77,9 @@ For the list of events logged to Stackdriver for the Import Adapter, see [here](
 
 The monitored resource is configured as k8s_container, with values set from a combination of environment variables configured via Downward API (pod name, pod namespace and container name) and GCP Metadata (project id, cluster name and location). Defaults to the global resource, if k8s_container can't be configured.
 
-Relevant part of yaml configuration:
+The following configuration needs to be added to the dicom-adapter.yaml file to configure the 
+stackdriver monitoring resource. Please see the (link)[Deployment using Kubernetes] section 
+for more information.
 ```yaml
 env:
 - name: ENV_POD_NAME
--- a/export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreSender.java
+++ b/export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreSender.java
@@ -80,7 +80,7 @@ public class CStoreSender implements DicomSender {
     }
 
     CountingInputStream countingStream = new CountingInputStream(part.getInputStream());
-    DicomClient.connectAndCstore(sopClassUid, sopInstanceUid, transferSyntaxUid, countingStream,
+    DicomClient.connectAndCstore(sopClassUid, sopInstanceUid, countingStream,
         applicationEntity, dimsePeerAET, dimsePeerIP, dimsePeerPort);
     MonitoringService.addEvent(Event.BYTES, countingStream.getCount());
   }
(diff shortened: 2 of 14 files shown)


import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+    private CMoveTask(Association as, PresentationContext pc,
+        Attributes cmd, Attributes keys) {
+      super(as, pc, cmd);
+
+      this.keys = keys;
+    }
+
+    @Override
+    public void run() {
+      List<String> failedInstanceUids = new ArrayList<>();
```
danielbeaudreau: Considering how in http://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.4.2.html Failed SOP Instance UID List  is all the SOP instances for which the operation failed, (and it's even supposed to be returned for a Refused status), I think the interpretation is that this list contains instance UIDs for every instance which did not complete processing. In the code we add instances incrementally to this list if they fail processing, but I think instead we should start with a set of all SOP instance uids returned by qido, and gradually remove them as CSTOREs complete. That way in the event of a cancel or an exception, the list contains all UIDs which did not complete processing.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AetDictionary.java
```diff
@@ -0,0 +1,100 @@
...
+      this.port = port;
+    }
+
+    @Override
+    public String toString() {
+      return "Aet{" +
+          "name='" + name + '\'' +
+          ", host='" + host + '\'' +
+          ", port=" + port +
+          '}';
```
danielbeaudreau: Do you need a \' after port?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AetDictionary.java
```diff
@@ -0,0 +1,100 @@
...
+import java.nio.file.Paths;
+import java.util.HashMap;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AetDictionary {
+
+  private static final String ENV_AETS_JSON = "ENV_AETS_JSON";
```
danielbeaudreau: We should add documentation to the README for this environment variable. (I believe the CMOVE documentation was in the original PR, but missing here)

danielbeaudreau:
Hi Egor, can you please also (in a separate PR) add an integration test for CFIND now that it's merged? Thanks

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+          }
+
+          JSONObject instanceJson = (JSONObject) instance;
+          String studyUid = AttributesUtil.getTagValue(instanceJson,
+              TagUtils.toHexString(Tag.StudyInstanceUID));
+          String seriesUid = AttributesUtil.getTagValue(instanceJson,
+              TagUtils.toHexString(Tag.SeriesInstanceUID));
+          String instanceUid = AttributesUtil.getTagValue(instanceJson,
+              TagUtils.toHexString(Tag.SOPInstanceUID));
+          String classUid = AttributesUtil.getTagValue(instanceJson,
```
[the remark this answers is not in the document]
danielbeaudreau: Would this cause an error on CSTORE for when TS or SOP ClassUID are missing? Alternatively we can just call getTagValue for both, since that throws an exception if the tag is not found.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+              TagUtils.toHexString(Tag.SOPClassUID));
+          String tsuid = AttributesUtil.getTagValueOrNull(instanceJson,
+              TagUtils.toHexString(Tag.TransferSyntaxUID));
+          if (tsuid == null) {
```
danielbeaudreau: Should we just throw an error if transfer syntax of the file is unknown? What if it's not ExplicitVRLittleEndian?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        JSONArray qidoResult;
+        try {
+          qidoResult = dicomWebClient.qidoRs(qidoPath);
+          if (qidoResult == null || qidoResult.length() == 0) {
+            throw new IDicomWebClient.DicomWebException("No instances to move");
+          }
+        } catch (IDicomWebClient.DicomWebException e) {
+          log.error("Failed QidoRs for CMove", e);
```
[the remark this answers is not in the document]
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+          log.info("CMove QidoPath: " + qidoPath);
+        } catch (DicomServiceException e) {
+          log.error("CMove QidoPath error");
+          sendErrorResponse(e.getStatus(), e.getMessage(), e.getDataset(), null);
+          return;
+        }
+
+        JSONArray qidoResult;
+        try {
+          qidoResult = dicomWebClient.qidoRs(qidoPath);
```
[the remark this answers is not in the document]
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+          }
+
+          remainingInstances--;
+        }
+
+        if (failedInstanceUids.isEmpty()) {
+          as.tryWriteDimseRSP(pc, Commands.mkCMoveRSP(cmd, Status.Success));
+        } else {
+          int status = successfullInstances > 0 ?
+              Status.OneOrMoreFailures : Status.UnableToPerformSubOperations;
```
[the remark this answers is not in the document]
red1408: From reading http://dicom.nema.org/medical/dicom/current/output/chtml/part04/sect_C.4.2.3.html#sect_C.4.2.3.1 both seem equally valid.
DCM4che chose to return UnableToPerformSubOperations in same situation https://dcm4che.atlassian.net/browse/DCMEE-30 . So at least I'm not the only one to pick this code.

To me UnableToPerformSubOperations seems to be more clear indication that problem is with sub operations rather than c-move as whole, but I can change to more generic status if you think it would be better.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -0,0 +1,240 @@
...
+
+    private void sendErrorResponse(int status, String message) {
+      sendErrorResponse(status, message, null, null);
+    }
+
+    private void sendErrorResponse(int status, List<String> failedInstanceUids) {
+      sendErrorResponse(status, null, null, failedInstanceUids);
+    }
+
+    private void sendErrorResponse(int status, String message, Attributes dataset,
```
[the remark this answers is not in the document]
red1408: it is on line 97:

> log.error("CMove QidoPath error");
> sendErrorResponse(e.getStatus(), e.getMessage(), e.getDataset(), null);

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AetDictionary.java
red1408: This is auto-generated toString(), seems that Idea chose to use ' instead of " around } because it's only one symbol. Changed to " for consistency.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AetDictionary.java
red1408: Right, restored relevant part of readme.

red1408:
I think it would be more convenient to test c-find and later c-move within same cloud build test (as extra steps), to reuse gradle/maven build steps which take 99% or so of time.
Or should I make them completely separate anyway?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
red1408: I think there is important distinction between not trying a sub operation and actually failing it.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
red1408: DCM4che QIDO-RS didn't return it for some reason, this default worked. 
However, I've just run some tests against cloud healthcare api and while I didn't see it not return TransferSyntaxUID, returned TransferSyntaxUID was wrong in some cases (= did not match metadata in actual file transferred via stow-rs, which leads to silent failure whens client tried to download these).
Since QIDO-RS value proved to be untrustworthy, I've changed code to use TransferSyntaxUID from file transferred via stow-rs. Seems to work without problems now.

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreSender.java
```diff
@@ -80,7 +80,7 @@ public void send(PubsubMessage message) throws Exception {
     }
 
     CountingInputStream countingStream = new CountingInputStream(part.getInputStream());
```
danielbeaudreau: Can you please remove the old logic in this file which gets transferSyntaxUid (and passes it as an include field)?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
danielbeaudreau: Hmm, it should always be accurate in the CHC API since it's derived from the uploaded file. But the new way of doing it looks good to me.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
danielbeaudreau: Ok, fair enough.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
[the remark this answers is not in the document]
danielbeaudreau: Ok, if that's what dcm4che does then this seems fine as is.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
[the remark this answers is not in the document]
danielbeaudreau: Ah, sorry I missed that. But correct me if I'm wrong, it doesn't seem like dataset would ever get set in that code path? (at least based on the version of DicomServiceException.java I'm looking at)

danielbeaudreau:
Ok, sounds good on waiting until cmove is submitted and sharing a build file for both tests.

README.md
```diff
@@ -9,6 +9,23 @@ two components, namely import and export adapter.
 The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests and passes the converted results back to the DIMSE client. The following requests are supported:
 - C-STORE to STOW-RS
 - C-FIND to QIDO-RS
+- C-MOVE uses QIDO-RS to determine instances to be transfered, then (per instance) executes WADO-RS to obtain instance data stream and passes it to C-STORE (to C-MOVE destination).
```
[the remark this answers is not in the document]
red1408: Sure

README.md
```diff
@@ -9,6 +9,23 @@ two components, namely import and export adapter.
 The Import Adapter converts incoming DIMSE requests to corresponding DICOMWeb requests and passes the converted results back to the DIMSE client. The following requests are supported:
 - C-STORE to STOW-RS
 - C-FIND to QIDO-RS
+- C-MOVE uses QIDO-RS to determine instances to be transfered, then (per instance) executes WADO-RS to obtain instance data stream and passes it to C-STORE (to C-MOVE destination). 
+
+AET resolution for C-MOVE is configured via AET dictionary json file ("--aet_dictionary" command line parameter or "ENV_AETS_JSON" environment variable). Format: JSON array of objects containing name, host and port.
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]

README.md
```diff
@@ -9,6 +9,23 @@ two components, namely import and export adapter.
...
+
+```shell
+kubectl create configmap aet-dictionary --from-file=AETs.json
+```
+
+Relevant part of yaml:
```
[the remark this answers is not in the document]

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreSender.java
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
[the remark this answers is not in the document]
red1408: Yeah, seems you are right, there is no code path to set it. Removed dataset param.

README.md
```diff
@@ -9,15 +9,41 @@ two components, namely import and export adapter.
...
+]
+```
 
 ```shell
 kubectl create configmap aet-dictionary --from-file=AETs.json
 ```
 
-Relevant part of yaml:
+The following configuration needs to be added to the dicom-adapter.yaml file to use CMOVE. 
+Please see the (link)[Deployment using Kubernetes] section for more information.
```
danielbeaudreau: Sorry for not being more clear, I meant have "Deployment using Kubernetes" link to that section here and in the stackdriver section.

README.md
```diff
@@ -9,6 +9,49 @@ two components, namely import and export adapter.
...
+	{
+		"name": "GINKGO_001",
+		"host": "some.domain.name.com",
+		"port": 6666
+	},
+	...
+]
+```
+
+```shell
```
danielbeaudreau: Can we move this to below "The following configuration needs to be added to the dicom-adapter.yaml file to use CMOVE. 
Please see the Deployment using Kubernetes section for more information." line?

danielbeaudreau:
Last two readme changes, then I will merge this. Thanks!

README.md
danielbeaudreau: I believe the format will be like [Deployment using Kubernetes](#deployment-using-kubernetes)

danielbeaudreau:
One more thing I forgot, can we replace the names in the json example (weasis, aeskulap, etc) with more generic names DEVICE_A, DEVICE_B to avoid naming other products in the readme? (And we can probably use just 2 examples to be more concise) Thanks!

README.md

README.md
red1408: Ok changed. It was already kind of highlighted when viewed locally in Idea, so I expected it to just work.

PR #18, opened 2019-07-16
4th unauthorized

red1408:
https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/issues/15

### DIFF
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
@@ -16,6 +16,7 @@ package com.google.cloud.healthcare;
 
 import com.github.danieln.multipart.MultipartInput;
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpMediaType;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
@@ -104,11 +105,11 @@ public class DicomWebClient implements IDicomWebClient {
    */
   public void stowRs(String path, InputStream in) throws IDicomWebClient.DicomWebException {
     GenericUrl url = new GenericUrl(serviceUrlPrefix + "/" + path);
-    MultipartContent content = new MultipartContent();
 
     // DICOM "Type" parameter:
     // http://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_6.6.1.1.1
-    content.setMediaType(content.getMediaType().setParameter("type", "\"application/dicom\""));
+    MultipartContent content = new MultipartContent();
+    content.setMediaType(new HttpMediaType("multipart/related; type=\"application/dicom\""));
     content.setBoundary(UUID.randomUUID().toString());
     InputStreamContent dicomStream = new InputStreamContent("application/dicom", in);
     content.addPart(new MultipartContent.Part(dicomStream));
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
@@ -19,6 +19,7 @@ import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.api.client.http.InputStreamContent;
 import com.google.api.client.http.MultipartContent;
@@ -28,6 +29,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.UUID;
 import javax.inject.Inject;
+import org.dcm4che3.net.Status;
 import org.json.JSONArray;
 
 /**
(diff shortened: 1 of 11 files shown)


dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
```diff
@@ -78,17 +80,26 @@ public JSONArray qidoRs(String path) throws IDicomWebClient.DicomWebException {
...
-            String.format(
-                "QidoRs: %d, %s", httpResponse.getStatusCode(), httpResponse.getStatusMessage()));
-      }
       // dcm4che server can return 204 responses.
       if (httpResponse.getStatusCode() == HttpStatusCodes.STATUS_CODE_NO_CONTENT) {
         return new JSONArray();
       }
       return new JSONArray(
           CharStreams.toString(new InputStreamReader(httpResponse.getContent(), "UTF-8")));
+    } catch (HttpResponseException e) {
```
danielbeaudreau: It would be nice to have this mapping logic in a function which for all three methods (wado, qido, and stow) use. Then we only have to worry about this mapping in one place, rather than having to make changes throughout the codebase when we want to update it. So all three of the methods would call into a function that maps the HTTP status to DIMSE status. And then we can migrate the CSTORE sender to use the DicomWebClient. What do you think?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -92,13 +92,16 @@ protected void store(
         HttpRequest httpRequest = requestFactory.buildPostRequest(url, content);
         httpRequest.execute();
       } catch (IOException e) {
-        if (e instanceof HttpResponseException &&
-            ((HttpResponseException) e).getStatusCode() ==
-                HttpStatusCodes.STATUS_CODE_SERVICE_UNAVAILABLE) {
-          throw new DicomServiceException(Status.OutOfResources, e);
-        } else {
-          throw e;
+        if (e instanceof HttpResponseException) {
```
danielbeaudreau: Can we use DicomWebClient instead here for making the stow request, and make changes so that STOW maps these status codes in the DicomWebClient?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -157,6 +151,10 @@ public void run() {
       } catch (CancellationException | InterruptedException e) {
         log.info("Canceled CMove", e);
         sendErrorResponse(Status.Cancel, failedInstanceUids);
+      } catch (IDicomWebClient.DicomWebException e) {
+        MonitoringService.addEvent(Event.CMOVE_QIDORS_ERROR);
```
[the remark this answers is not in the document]
red1408: cstoreSender.cstore has lower level catch block around it, so this wouldn't happen. Still, moved catch for qido-rs block back to separate try to make these more distinct.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
red1408: Yes, makes sense.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Ok

integration_test/run_adapter_and_wait.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # apiStage project location dataset store adapter_port finish_port
-gradle run -Dexec.args='--dimse_aet=IMPORTADAPTER --dimse_port='$6' --dicomweb_addr=https://healthcare.googleapis.com/'$1'/projects/'$2'/locations/'$3'/datasets/'$4'/dicomStores/'$5'/dicomWeb --dicomweb_stow_path=/studies --verbose' &
+gradle run -Dexec.args='--dimse_aet=IMPORTADAPTER --dimse_port='$6' --dicomweb_addr=https://healthcare.googleapis.com/'$1'/projects/'$2'/locations/'$3'/datasets/'$4'/dicomStores/'$5'/dicomWeb --dicomweb_stow_path=studies --verbose' &
```
danielbeaudreau: We don't want to break existing clients which specify the path starting with "/". Let's change the code of the dicomweb client stowrs method to handle all cases (base url has trailing /, path has a / at the beginning, etc). We can do this by trimming them off of both the base url and path if they exist.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
```diff
@@ -135,6 +120,10 @@ public void run() {
         log.info("Canceled CFind", e);
         MonitoringService.addEvent(Event.CFIND_CANCEL);
         as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.Cancel));
+      } catch (IDicomWebClient.DicomWebException e) {
+        log.error("CFind qido-rs error:", e);
```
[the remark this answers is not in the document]
red1408: ok

integration_test/run_adapter_and_wait.sh
red1408: Do we actually need dicomweb_stow_path parameter at all? I mean it's present as legacy, but I'm not sure why it was needed in the 1st place. It's always 'studies' anyway (with or without \).

The only other option in stow-rs standard is '/studies/{study}', and difference is that this produces errors on mismatch between url study uid and instance study uid ( http://dicom.nema.org/dicom/2013/output/chtml/part18/sect_6.6.html#sect_6.6.1.1 ). Does not seem particularly important.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -81,38 +72,22 @@ protected void store(
...
-      } catch (IOException e) {
-        if (e instanceof HttpResponseException &&
-            ((HttpResponseException) e).getStatusCode() ==
-                HttpStatusCodes.STATUS_CODE_SERVICE_UNAVAILABLE) {
-          throw new DicomServiceException(Status.OutOfResources, e);
-        } else {
-          throw e;
-        }
-      }
+      dicomWebClient.stowRs( path, inBuffer);
```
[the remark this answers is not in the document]
red1408: Fixed this line. Well, I set 'Reformat code' before commit as default in my local Idea options, but I'm not sure how I could implement reformat or check as cloud build trigger. Any suggestions?

integration_test/run_adapter_and_wait.sh
danielbeaudreau: It's a good point. Ideally we should just have one dicomweb address parameter which goes up to /dicomWeb, which is used by CFIND, CMOVE, and CSTORE.

It's important to keep backwards compatibility here for now. I spoke to Dusan about it and I think the following plan makes sense:
1. (this PR) Support existing clients which specify both paths separately, and handle all cases of where the "/" character is present as I mentioned in the first comment. The reason we can't get rid of the stow path param and hardcode to /studies is that current clients don't specify the full dicom web address in the dicom web address flag. From our readme:
- "--dicomweb_addr=https://healthcare.googleapis.com/v1beta1"
            - "--dicomweb_stow_path=/projects/myproject/locations/us-central1/datasets/mydataset/dicomStores/mydicomstore/dicomWeb/studies"
Note the stow_path is not just /studies here.

2. (follow up PR) Add a new flag for dicom web address which is mandatory to use CFIND and CMOVE, which is the dicomweb path up to /dicomWeb. We can use only this for CFIND and CMOVE. For CSTORE since we have legacy clients we should support both ways of specifying the path for a while. Note that CFIND and CMOVE won't work with the current way of specifying the path in the readme file above, because the complete path is not part of the dicomweb_addr param.

3. (same follow up PR) Mark the old stow path flag as deprecated in a comment, and explain how to migrate to the new way of specifying the flag in the readme, as well as updating the example in the readme.

4. Eventually we can remove the stow path. 

Does this sound good to you?

red1408: Ok, makes sense. Added  "/" trimming.

PR #25, opened 2019-08-02
7th better integration test

red1408:
Planning to add storage commitment test too, but decided to create pull request now to start review earlier.

EDIT: added simple commitment test.


integration_test/cloudbuild-integration-test.yaml
```diff
@@ -89,13 +161,32 @@
...
+- name: 'busybox:latest'
+  id: check-find-diff
+  args:
+  - diff
+  - 'integration_test/findscu-result1.xml'
+  - 'integration_test/findscu-expected.xml'
+  waitFor:
+  - run-find-scu
+
+# storescp adds 1 extra tag
```
danielbeaudreau: Just wondering, which tag does it add?

integration_test/run-adapter-and-wait.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
-# apiStage project location dataset store adapter_port finish_port
-gradle run -Dexec.args='--dimse_aet=IMPORTADAPTER --dimse_port='$6' --dicomweb_addr=https://healthcare.googleapis.com/'$1'/projects/'$2'/locations/'$3'/datasets/'$4'/dicomStores/'$5'/dicomWeb --dicomweb_stow_path=/studies --verbose' &
+# apiStage project location dataset store adapter_port finish_port storescp_host store_scp_port
```
danielbeaudreau: Missing a comment for storage management host / port

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
```diff
@@ -68,6 +68,12 @@
   )
   String aetDictionaryPath = "";
 
+  @Parameter(
```
danielbeaudreau: I'd prefer to keep this simple and not potentially confuse users by having two flags for the same parameter. Can we just use the existing AET dictionary as a file for this test?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -47,13 +55,23 @@
   id: wait-for-adapter
   args:
   - ash
-  - integration_test/wait_for_port.sh
+  - integration_test/wait-for-port.sh
   - ${_ADAPTER_RUN_STEP}
   - ${_ADAPTER_PORT}
   waitFor:
-  - build-store-scu
+  - build-tools
```
danielbeaudreau: Shouldn't this wait for adapter-main instead of build-tools? (not that it matters too much)

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -47,13 +55,23 @@
...
 
+- name: 'busybox:latest'
+  id: wait-for-storescp
+  args:
+  - ash
+  - integration_test/wait-for-port.sh
+  - ${_STORE_SCP_RUN_STEP}
+  - ${_STORE_SCP_PORT}
+  waitFor:
+  - build-tools
```
danielbeaudreau: Shouldn't this wait for store-scp instead?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -68,6 +86,50 @@
...
+  - mvn
+  - 'exec:java'
+  - '-Dexec.mainClass=org.dcm4che3.tool.findscu.FindSCU'
+  - '-Dexec.args=-c IMPORTADAPTER@${_ADAPTER_RUN_STEP}:${_ADAPTER_PORT} -L IMAGE -X --out-cat --out-file findscu-result.xml --out-dir ../../../integration_test/'
+  dir: "dcm4che/dcm4che-tool/dcm4che-tool-findscu"
+  waitFor:
+  - run-store-scu
+
+- name: "maven:3.6.0-jdk-8"
+  id: run-commitment-scu
```
danielbeaudreau: How do we know if storage commitment SCU returned a success result? Is this captured in the return value of the mvn exec command? (0 for success, non-0 for failure?) If so, can you please add a comment explaining this.

integration_test/cloudbuild-integration-test.yaml
red1408: Ok, added comment.

integration_test/run-adapter-and-wait.sh
red1408: Ok, added

integration_test/cloudbuild-integration-test.yaml
red1408: 'waitFor:' means waiting for docker container running the step to close with success result. That's not what I need, I need adapter and storescp up and ready to execute queries.

integration_test/cloudbuild-integration-test.yaml
red1408: Same as above.

integration_test/cloudbuild-integration-test.yaml
red1408: Commitment SCU waits for n-event-report after n-action request returns success, if it doesn't get one within cloud build step timeout of 30 seconds, step fails.

In theory, Commitment SCU would actually return 0 (success), if n-action request fails (malformed commitment request or unknown destination AET), but 1st is impossible (we are using known correct client), while 2nd would be a bug in integration itself.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Passing it this way also would also simplify gke/gce setup (no need to create configmap for gke/configure volume for gce), at least when AET dictionary is small enough.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -111,7 +205,9 @@
   - "--location=${_LOCATION}"  
   - "--quiet"
   waitFor:
-  - diff-dcm
+  - check-store-curl
+  - run-find-scu
+  - run-move-scu
```
danielbeaudreau: Let's add run-commitment-scu here for consistency?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Ok

integration_test/cloudbuild-integration-test.yaml
red1408: Commitment SCU does dump what it receives by default. Problem is, n-event-report contains random transaction id, so I can't compare them using diff. While regex tools don't seem to work all that well on binary files.

integration_test/cloudbuild-integration-test.yaml
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
```diff
@@ -68,6 +68,13 @@
...
+      names = {"--aet_dictionary_inline"},
+      description = "Json array containing aet definitions (name/host/port per element)."
+          + "Only one of aet_dictionary and aet_dictionary_inline needs to be specified."
```
danielbeaudreau: Missing a space between sentences.

integration_test/cloudbuild-integration-test.yaml
danielbeaudreau: But, if a server failure occurs for storage commit, the integration test won't fail since an n-event-response will be sent anyway.

What if we install the dcm2xml tool (https://github.com/dcm4che/dcm4che/tree/master/dcm4che-tool/dcm4che-tool-dcm2xml) and use this to dump the XML? Then we can remove the transaction ID using a tool like sed. Will this work? I'm not sure if dcm2xml will work on a DIMSE response.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Ok

integration_test/cloudbuild-integration-test.yaml
red1408: Added actual diff check.

I forgot to bring up one issue though (mentioned it only in git comment): api does (I'm pretty sure incorrectly) return  "generic::invalid_argument: SOPClassUID is not a supported instance or study or series level attribute" on qido-rs queries used by commitment service.
Most of my testing is done with local dcm4che, where such queries work.

So current commitment test just validates that api still doesn't work.

danielbeaudreau:
I see. I filed issue #28 to address this, please take a look

Issue #28, opened 2019-08-09
Modify storage commitment adapter to query only on sop instance uid (temporarily)


danielbeaudreau:
When it's supported on the API side, we can again query on SopClassUID as well.

PR #30, opened 2019-08-12
10th temp commitment instanceuid only

red1408:
[#28] Modify storage commitment adapter to query only on sop instance uid (temporarily)

Needs 9th pull request.


import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/StorageCommitmentService.java
```diff
@@ -149,8 +149,10 @@ public void run() {
             attrsItem.getString(Tag.ReferencedSOPClassUID));
         queryAttributes
             .setString(Tag.SOPInstanceUID, VR.UI, cmtItem.getInstanceUid());
-        queryAttributes
-            .setString(Tag.SOPClassUID, VR.UI, cmtItem.getClassUid());
+        // current (v1beta1) healthcare api wrongly returns
+        // "SOPClassUID is not a supported instance or study or series level attribute"
+        //queryAttributes
```
danielbeaudreau: I'd prefer deleting the code for now, and we can use commit history to add it back once the issue is resolved on the server side.

red1408: Ok

danielbeaudreau:
Thanks, can you please rebase this as well?  (I guess all the rebasing is getting a bit problematic, but this PR doesn't remove the duplicate test script file, so it's necessary here anyway)

red1408:
Done.

Issue #19, opened 2019-07-19
Deprecate stow path, introduce a new parameter for the dicom web address


danielbeaudreau:
From https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/pull/18:

Add a new flag for dicom web address which is mandatory to use CFIND and CMOVE, which is the dicomweb path up to /dicomWeb. We can use only this for CFIND and CMOVE. For CSTORE since we have legacy clients we should support both ways of specifying the path for a while. Note that CFIND and CMOVE won't work with the current way of specifying the path in the readme file above, because the complete path is not part of the dicomweb_addr param.

Mark the old stow path flag as deprecated in a comment, and explain how to migrate to the new way of specifying the flag in the readme, as well as updating the example in the readme.

Eventually we can remove the stow path.

red1408:
What do you think would be a good name for new full dicomweb address parameter? Something like "--dicomweb_addr_full" just looks superfluous.

danielbeaudreau:
Maybe we can just use --dicomweb_address and update our readme to use the new param. The old param can just exist in the flags file for now, and we can handle it on the server side for now, but all new users will specify the --dicomweb_address.

danielbeaudreau:
This is mainly done, however we'd like to also deprecate the peer addr/stowpath in the export adapter.

Issue #5, opened 2019-06-07
Propagate API Error Message on Store Failure


dzelemba:
If you bring up the adaptor, but haven't configured the API target properly (e.g. a typo in the name or missing permissions), you'll just get a generic error message when you try and push an image through. Instead, we should return the error message that the API returns.

red1408:
After looking into it: problem is, situation like this fails at low level while sending data stream. There is no http response or http exception involved. I just don't receive a response.

```
org.dcm4che3.net.service.DicomServiceException: java.io.IOException: Error writing request body to server( status = 272 )
	at com.google.cloud.healthcare.imaging.dicomadapter.CStoreService.store(CStoreService.java:84)
	at org.dcm4che3.net.service.BasicCStoreSCP.onDimseRQ(BasicCStoreSCP.java:72)
	at org.dcm4che3.net.service.DicomServiceRegistry.onDimseRQ(DicomServiceRegistry.java:86)
	at org.dcm4che3.net.ApplicationEntity.onDimseRQ(ApplicationEntity.java:485)
	at org.dcm4che3.net.Association.onDimseRQ(Association.java:650)
	at org.dcm4che3.net.PDUDecoder.decodeDIMSE(PDUDecoder.java:459)
	at org.dcm4che3.net.Association.handlePDataTF(Association.java:634)
	at org.dcm4che3.net.State$4.onPDataTF(State.java:103)
	at org.dcm4che3.net.Association.onPDataTF(Association.java:630)
	at org.dcm4che3.net.PDUDecoder.nextPDU(PDUDecoder.java:177)
	at org.dcm4che3.net.Association$2.run(Association.java:478)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: com.google.cloud.healthcare.IDicomWebClient$DicomWebException: java.io.IOException: Error writing request body to server
	at com.google.cloud.healthcare.DicomWebClient.stowRs(DicomWebClient.java:128)
	at com.google.cloud.healthcare.imaging.dicomadapter.CStoreService.store(CStoreService.java:75)
	... 13 more
Caused by: java.io.IOException: Error writing request body to server
[...]

PR #31, opened 2019-08-20
[#19] Deprecate peer_dicomweb_addr peer_dicomweb_stow_path

red1408:


red1408:
Had to bake into https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/pull/34
Got some errors while testing export adapter with old stowrs implementation (unclear what, for same reason as #5), and while I potentially could fix them, I don't see much point since next step is to change stowrs implementation to jetty (which doesn't encounter any errors in same situation).

Issue #13, opened 2019-06-28
Add versioning to containers 


danielbeaudreau:
Please see comment below for a description of this task.

danielbeaudreau:
The plan is to have a cloud build trigger which runs after each commit to master:

- Runs unit tests to ensure they pass
- Automatically builds the docker containers
- Runs integration tests against the docker containers
- If tests pass, publish a new image to a gcr.io location tagged with the commit ID of the commit 
  which was built. Note that the gcr.io location will be new (different from existing one) so we don’t 
  risk breaking any users. We’ll use the following new URLS:
  gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-import:{version id}
  gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-export:{version id}
- Applies a version number to the container of the format vX.Y.Z where:
  We increment X manually on breaking changes
  We increment Y manually on new feature additions
  Cloud build script auto-increments Z

danielbeaudreau:
Hi Egor, this is ready to be worked on now.

We will need to make some configuration changes in the oss-test project for this to work. I can make these changes once we have the basic cloud build file in a PR.

danielbeaudreau:
This is working now. 
https://pantheon.corp.google.com/gcr/images/cloud-healthcare-containers/GLOBAL/healthcare-api-dicom-dicomweb-adapter-import?gcrImageListsize=30

pavertomato:
> This is working now.
> https://pantheon.corp.google.com/gcr/images/cloud-healthcare-containers/GLOBAL/healthcare-api-dicom-dicomweb-adapter-import?gcrImageListsize=30

this link is available only for googlers:(

danielbeaudreau:
Woops, thanks for catching that.
This is the link:
gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-import

dzelemba:
Two small things left here:
  1) Cleanup README to use new container
  2) Add WIKI page describing release process.

pavertomato:
https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/wiki/Release-process
is updated

Issue #1, opened 2019-06-07
Setup Continuous Build


dzelemba:
Using Google Cloud Build.

PR #32, opened 2019-08-30
versioning for images

pavertomato:
[#13]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,11 @@ ext.downloadLicensesText = { buildDir ->
   new File(buildDir.toString() + '/docker/license').mkdirs()
   for (dep in parsedJson.dependencies) {
     def licenseFile = new File(buildDir.toString() + '/docker/license/' + dep.name)
-    licenseFile << new URL(dep.licenses[0].url).text
+    try {
+      licenseFile << new URL(dep.licenses[0].url).text
+    } catch (IOException e) {
+      logger.debug("can't download license file:" + dep.licenses[0].url)
+    }
   }
 }
 
--- a/integration_test/build-adapters.sh
+++ b/integration_test/build-adapters.sh
@@ -12,10 +12,14 @@ build_adapter () {
   gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$version.$postfix
   gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$2
 }
-
+apt-get -qq update
+apt-get install -y gnupg2
+# we need gcloud and 5.6-jdk11 in same image next 4 lines adds google's package storage and installs gcloud by this manual https://cloud.google.com/sdk/docs/quickstart-debian-ubuntu
+echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
 apt-get -qq update
 apt-get install google-cloud-sdk -y
-apt-get -qq install -y gradle
+
 gradle build
 cd ./import
 build_adapter import $2
--- a/integration_test/build-adapters.sh
+++ b/integration_test/build-adapters.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# imageProject commitShortSha
+
+base_name="gcr.io/${1}/healthcare-api-dicom-dicomweb-adapter-"
+
+build_adapter () {
+  gradle dockerBuildImage
+  adapter_name=${base_name}${1}
+  version=$(gradle printVersion  | grep -Eo "[0-9]{1,3}\.[0-9]{1,3}" | head -1)
+  postfix=$(gcloud container images list-tags "${adapter_name}" --sort-by=~timestamp | grep " $version" |  grep -Eo "([0-9]{1,3}\.){2}[0-9]{1,3}" | head -1 | grep -Eo "[0-9]{1,3}" | tail -1)
+  postfix=$(($postfix+1))
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$version.$postfix
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$2
+}
+
+apt-get -qq update
+apt-get -qq install -y gradle
+gradle build
+cd ./importa
+build_adapter import $2
+cd ../export
+build_adapter export $2
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -239,7 +239,7 @@
   id: publish-image
   args: [ 'bash', 'integration_test/publish-images.sh', '${_IMAGEPROJECT}', "${_PUBLISH}"]
   waitFor: ['delete-dicom-store']
-
+timeout: 1800s
 substitutions:
   _VERSION: v1beta1
   _PROJECT: gcp-healthcare-oss-test
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -31,12 +31,19 @@
   - ${_STORE_SCP_PORT}
   - ${_CLOSE_STORE_SCP_PORT}
 
-# run adapter, close when connection established on _CLOSE_ADAPTER_PORT
-- name: 'gradle:5.4.1-jdk8'
-  id: adapter-main
-  args: [ 'bash', '../integration_test/run-adapter-and-wait.sh', '${_VERSION}', '${_PROJECT}', '${_LOCATION}', '${_DATASET}', '${_STORE_NAME}', '${_ADAPTER_PORT}', '${_CLOSE_ADAPTER_PORT}', '${_STORE_SCP_RUN_STEP}', '${_STORE_SCP_PORT}', '${_COMMITMENT_SCU_STEP}', '${_COMMITMENT_SCU_PORT}']
-  dir: import
+- name: 'google/cloud-sdk:250.0.0'
+  id: adapter-build
+  args: [ 'bash', './integration_test/build-adapters.sh', '${_IMAGEPROJECT}', '${SHORT_SHA}']
   waitFor: ['-']
+
+# run adapter
+
+- name: 'gcr.io/${_IMAGEPROJECT}/healthcare-api-dicom-dicomweb-adapter-import:${SHORT_SHA}'
+  id: adapter-main
+  entrypoint: 'bash'
+  args: [ '-c',
+          "/import/bin/import --dimse_aet=IMPORTADAPTER --dimse_port=${_ADAPTER_PORT} --dicomweb_address=https://healthcare.googleapis.com/${_VERSION}/projects/${_PROJECT}/locations/${_LOCATION}/datasets/${_DATASET}/dicomStores/${_STORE_NAME}/dicomWeb --aet_dictionary_inline=\"[{\"name\":\"STORESCP\",\"host\":${_STORE_SCP_RUN_STEP},\"port\":${_STORE_SCP_PORT}},{\"name\":\"STGCMTSCU\",\"host\":${_COMMITMENT_SCU_STEP},\"port\":${_COMMITMENT_SCU_PORT}},]\" --verbose & trap \"exit 0\" SIGTERM; while true; do sleep 1; done"]
+  waitFor: ['adapter-build']
 
 # create dataset if not exists, re-create store (delete if exists, then create)
 - name: 'google/cloud-sdk:250.0.0'
@@ -61,6 +68,7 @@
   timeout: 5m
   waitFor:
   - build-tools
+  - adapter-build
   - setup-dataset-and-dicom-store
 
 - name: 'busybox:latest'
@@ -138,15 +146,14 @@
   - run-commitment-scu
   - wait-for-storescp
 
-- name: 'busybox:latest'
-  id: close-adapter
+- name: 'gcr.io/cloud-builders/docker'
+  id:  close-adapter
+  entrypoint: bash
   args:
-  - nc
-  - '-z'
-  - ${_ADAPTER_RUN_STEP}
-  - ${_CLOSE_ADAPTER_PORT}
-  waitFor:
-  - run-move-scu
+      - './integration_test/stop-image.sh'
+      - '${_IMAGEPROJECT}'
+      - '${SHORT_SHA}'
+  waitFor: ['run-move-scu']
 
 - name: 'busybox:latest'
   id: close-store-scp
@@ -220,7 +227,7 @@
   - ${_STORE_NAME}
   - "--project=${_PROJECT}"
   - "--dataset=${_DATASET}"
-  - "--location=${_LOCATION}"  
+  - "--location=${_LOCATION}"
   - "--quiet"
   waitFor:
   - check-store-curl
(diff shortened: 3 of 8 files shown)


pavertomato:
> denied: Token exchange failed for project 'cloud-healthcare-containers'. Caller does not have permission 'storage.buckets.get'. To configure permissions, follow instructions at: https://cloud.google.com/container-registry/docs/access-control
The push refers to repository [gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-export]

seems like gcp-healthcare-oss-test  don't have permissions to push into cloud-healthcare-containers as specified in #13

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+base_name="gcr.io/${1}/healthcare-api-dicom-dicomweb-adapter-"
+
+build_adapter () {
+  adapter_name=${base_name}${1}
+  version=$(cat build.gradle | grep "version = '"| grep -Eo "[0-9]{1,3}\.[0-9]{1,3}")
```
danielbeaudreau: Can we move the version from the gradle build file into a standalone file called VERSION? This file can contain the first two version numbers: X.Y
There are gradle commands which can be used to read the file. That way it's easy for someone to update/check the version, rather than having to look into the gradle.build files.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -220,27 +227,32 @@
...
-  - "--location=${_LOCATION}"  
+  - "--location=${_LOCATION}"
   - "--quiet"
   waitFor:
   - check-store-curl
   - run-find-scu
   - run-move-scu
   - run-commitment-scu
 
+- name: 'google/cloud-sdk:250.0.0'
```
danielbeaudreau: I think we want publishing images to be done post-submit rather than as a pre-submit. For that reason I think it's better to have the publish command as a separate .yaml file.

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,18 @@
...
+  postfix=$(gcloud container images list-tags "${adapter_name}" --sort-by=~tags | grep " $version" |  grep -Eo "([0-9]{1,3}\.){2}[0-9]{1,3}" | head -1 | grep -Eo "[0-9]{1,3}" | tail -1)
```
danielbeaudreau: What if we bump the version from 0.1 to 0.2? Then if the most recent container is 0.1.8, we will write 0.2.9 rather than 0.2.0, right?

danielbeaudreau: I think the default order of ~timestamp would be better than by tag. Incrementing past 9 changes the last number to 10, this may cause errors in the order if it's alphabetical.

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,18 @@
...
+  postfix=$(($postfix+1))
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$version.$postfix
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:latest
```
danielbeaudreau: We're going to avoid pushing to :latest for now to prevent users from depending on a version which could break. This  reduces the risk of new changes.

integration_test/cloudbuild-integration-test.yaml
pavertomato: not actually clear:
for the moment images published only if integration tests is successful
if we move publishing to separate file we'll have to build images again and run tests again it'll make new yaml almost exactly the same
or it can be manually triggered, or by tag, in this case we may remove file with version

integration_test/build-adapters.sh
pavertomato: this expression filters string by $version(0.2 in your example) so postfix will be empty at line 7 and 1 at line 8, so result will be 0.2.1

integration_test/build-adapters.sh
pavertomato: we'll have to keep latest tag at build and tests steps, because this is default tag for image if tag is not specified, so i suggest to remove latest tag at publish step

integration_test/build-adapters.sh
pavertomato: thanks it's indeed alphabetical

integration_test/build-adapters.sh
pavertomato: >That way it's easy for someone to update/check the version

this is may be will make it easy for someone who is not familiar with gradle but definitely will make it confusing for someone who used to work with gradle, so it's controversial.
>There are gradle commands which can be used to read the file. 

i may suggest to add something like 

`
task printVersion {
  println project.version
}
`
but output won't be clear and we'll need to grep it as well
image

integration_test/cloudbuild-integration-test.yaml
pavertomato: >Can we define a variable $PUBLISH

this will requre manipulation with trigger, not sure if it's very handy
What do you think about publishing by tag? if we have tag in commit we'll publish image after integration tests, also it's removes doubts about versions - tag contains clear string

integration_test/build-adapters.sh
pavertomato: responded here https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/pull/32#discussion_r321152342

integration_test/build-adapters.sh
pavertomato: in cloudbuild yamls all variables resolved before any step is sterted
at step adapter-main we run freshly builded image, so we can't know exact version of it, if we left it empty then docker uses "latest" from local storage.
1 - we build image with tags: version, hash and latests
2 - tests
3 - we remove latests tag and publish image with tags: version, hash

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+base_name="gcr.io/${1}/healthcare-api-dicom-dicomweb-adapter-"
+
+build_adapter () {
+  adapter_name=${base_name}${1}
+  version=$(cat build.gradle | grep "version = '"| grep -Eo "[0-9]{1,3}\.[0-9]{1,3}")
+  postfix=$(gcloud container images list-tags "${adapter_name}" --sort-by=~timestamp | grep " $version" |  grep -Eo "([0-9]{1,3}\.){2}[0-9]{1,3}" | head -1 | grep -Eo "[0-9]{1,3}" | tail -1)
```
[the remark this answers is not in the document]
pavertomato: yes, i've tested it on our cloud build
also it can be proofed by running `echo $((""+1))`

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -31,13 +31,21 @@
   - ${_STORE_SCP_PORT}
   - ${_CLOSE_STORE_SCP_PORT}
 
-# run adapter, close when connection established on _CLOSE_ADAPTER_PORT
-- name: 'gradle:5.4.1-jdk8'
-  id: adapter-main
-  args: [ 'bash', '../integration_test/run-adapter-and-wait.sh', '${_VERSION}', '${_PROJECT}', '${_LOCATION}', '${_DATASET}', '${_STORE_NAME}', '${_ADAPTER_PORT}', '${_CLOSE_ADAPTER_PORT}', '${_STORE_SCP_RUN_STEP}', '${_STORE_SCP_PORT}', '${_COMMITMENT_SCU_STEP}', '${_COMMITMENT_SCU_PORT}']
+- name: 'google/cloud-sdk:250.0.0'
+  id: adapter-build
+  args: [ 'bash', '../integration_test/build-adapters.sh', '${_IMAGEPROJECT}', '${SHORT_SHA}']
```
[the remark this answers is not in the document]
pavertomato: done

build.gradle
```diff
@@ -27,7 +27,11 @@ ext.downloadLicensesText = { buildDir ->
   new File(buildDir.toString() + '/docker/license').mkdirs()
   for (dep in parsedJson.dependencies) {
     def licenseFile = new File(buildDir.toString() + '/docker/license/' + dep.name)
-    licenseFile << new URL(dep.licenses[0].url).text
+    try {
+      licenseFile << new URL(dep.licenses[0].url).text
+    } catch (IOException e) {
+      logger.debug("can't download licnese file:" + dep.licenses[0].url)
```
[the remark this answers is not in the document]
pavertomato: done

integration_test/cloudbuild-integration-test.yaml
danielbeaudreau: It would be very easy I think, since cloud build allows you to substitute variables for each trigger. We would simply set the publish variable to true for the post-submit run, and leave it as the default of false for the pre-submit.

If we publish in presubmits at all then we're publishing un-reviewed changes to the repo, right? Even if the commit ID is in the label instead of the version, users wouldn't be able to differentiate between a non-reviewed CL and a submitted CL.

integration_test/build-adapters.sh
danielbeaudreau: Instead of using the gcr.io/${_IMAGEPROJECT}/healthcare-api-dicom-dicomweb-adapter-import builder in the adapter-main step, can we instead run a bash command which runs the exact docker images we built in the build_adapter step? This would remove the problem of us picking up the :latest image from that gcr.io builder, if I understand correctly.

pavertomato: >can we instead run a bash command which runs the exact docker images

it was my original idea but it won't work because when we run docker inside of docker images is not authorized with cloud build service account
so we have 2 workarounds here - run image from yaml or store service accout file in gcs, download it and pass into image at start

danielbeaudreau: If you build the docker image with a tag, you should be able to execute that tag as the builder. An example: https://cloud.google.com/cloud-build/docs/configuring-builds/build-test-deploy-artifacts#writing_build_requests_without_pushing_images

So we should be able to execute the local image by using its tag, right?

integration_test/cloudbuild-integration-test.yaml
pavertomato: >It would be very easy I think

yeah, it's subjective

> we're publishing un-reviewed changes to the container repo, right?

yes

also i have idea how to simplify this process - we adopt git-flow for this repo and publish images only for updates in master branch, it means we'll publish only reviewed code
so for the moment we have 3 options

1. var in cloud build trigger
2. tag in commit  - same way as for https://github.com/GoogleCloudPlatform/wsi-to-dicom-converter/wiki/Release-process
3. changes in master - i think most clear way 

all of them quite simple to implement, please choose one

danielbeaudreau: I spoke to Dusan and he said we want to go with option #1, because the var in the cloud build trigger solves the issue of duplicating the YAML file, which I think we need to solve either way. This will also be essentially the same as option 3, as we are publishing a new container for each new commit to master - although we won't have separate dev and master branches.

integration_test/build-adapters.sh
pavertomato: done

integration_test/cloudbuild-integration-test.yaml
pavertomato: done

integration_test/build-adapters.sh
pavertomato: >An example: https://cloud.google.com/cloud-build/docs/configuring-builds/build-test-deploy-artifacts#writing_build_requests_without_pushing_images
i'm actually doing exact same thing, just note what "- name: 'analyzer'" not using any 'analyzer', it's using 'analyzer:latest' if we won't specify tag, so we have :latest image into local storage because we know tag of image only on step of building it

pavertomato: but i've forgot about commit hash, fixed

danielbeaudreau: Nice, thanks.

integration_test/publish-images.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
danielbeaudreau: Could you please add a comment line for the parameters to the scripts on the second line similar to run-adapter-and-wait? It will improve the readability of the scripts (publish-images and stop-image)

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -220,27 +227,33 @@
...
   - "--quiet"
   waitFor:
   - check-store-curl
   - run-find-scu
   - run-move-scu
   - run-commitment-scu
 
+- name: 'google/cloud-sdk:250.0.0'
+  id: publish-image
+  args: [ 'bash', 'integration_test/publish-images.sh', '${_IMAGEPROJECT}', "${_PUBLISH}"]
```
danielbeaudreau: I might be wrong, but I thought we needed to define substitutions in a substitutions map like here: https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions

pavertomato: it done at line 259 https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/pull/32/files/f7e916518509c3ea775afdbeb9f19b623e735906#diff-98e8aebae06d145f376ebd5cf5c14035R259
or you meant something else?

danielbeaudreau: Oh I see, sorry I thought it had to be in the same build step.

integration_test/publish-images.sh
pavertomato: done

danielbeaudreau: Thanks, also I think publish var is missing from publish-image script.

pavertomato: oh sorry
fixed

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,22 @@
...
+
+base_name="gcr.io/${1}/healthcare-api-dicom-dicomweb-adapter-"
+
+build_adapter () {
+  gradle dockerBuildImage
+  adapter_name=${base_name}${1}
+  version=$(gradle printVersion  | grep -Eo "[0-9]{1,3}\.[0-9]{1,3}" | head -1)
+  postfix=$(gcloud container images list-tags "${adapter_name}" --sort-by=~timestamp | grep " $version" |  grep -Eo "([0-9]{1,3}\.){2}[0-9]{1,3}" | head -1 | grep -Eo "[0-9]{1,3}" | tail -1)
+  postfix=$(($postfix+1))
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$version.$postfix
```
danielbeaudreau: One minor thing that came up from internal review, can we add the prefix "v" here? So instead of X.Y.Z, the containers will be labeled as vX.Y.Z. After that I'll merge this PR and configure the post submit build with the publish flag set to true.

danielbeaudreau:
Please fix the merge conflicts, then this can be merged. (and please disregard the previous email you may have gotten about adding a "v" prefix to the version numbers, it's ok as-is)

pavertomato:
> Please fix the merge conflicts, then this can be merged. (and please disregard the previous email you may have gotten about adding a "v" prefix to the version numbers, it's ok as-is)

done

integration_test/build-adapters.sh
```diff
@@ -0,0 +1,25 @@
...
+  gradle dockerBuildImage -Pdocker_tag=${adapter_name}:$2
+}
+apt-get -qq update
+apt-get install -y gnupg2
+echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
```
danielbeaudreau: Can you add a comment explaining what these lines do and why they're needed?

pavertomato: done, after recent update we've moved to java 11 so we need gcloud and gradle 5.6 with jdk11 in same image. standard ubuntu repo don't have java 11, gradle 5 and gcloud.
gcloud is smaller and packages hosted by google so i've decided to use gradle's image and install gcloud by this manual https://cloud.google.com/sdk/docs/quickstart-debian-ubuntu

PR #33, opened 2019-09-03
Stowrs error reporting and export adapter changes

red1408:
- Deprecate peer_dicomweb_addr peer_dicomweb_stow_path
- Jetty client http2 implementation of stow-rs used by c-store service and export adapter. Main difference - it can receive (error) response while transmitting data stream.
- updated to java 11. Initially because I tried to use native new java http client. Later left as is, since there seem to be no drawbacks.

Baked both pull requests into one because export adapter got broken pipe errors while trying to use old stowrs implementation (exactly same as #5), but when I tried to replace client with Jetty to actually see the error, it just worked. Doubt it's worth looking in too deeply, since that's what I was going to do next anyway.

Note: Jetty client implements only stow-rs (POST with binary data). I don't see particularly good reason to over-complicate GET requests with use of low level api.

EDIT: and it looks like I'll need to fix CLA compliance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,8 @@ containers in `dicom_adapter.yaml`. Modify the flags for your use case.
             - "--oauth_scopes=https://www.googleapis.com/auth/pubsub"
 ```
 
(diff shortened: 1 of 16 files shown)


red1408:
Re-created with fixed cla

PR #34, opened 2019-09-04
Stowrs error reporting and export adapter changes (fixed cla)

red1408:
- Deprecate peer_dicomweb_addr peer_dicomweb_stow_path
- Jetty client http2 implementation of stow-rs used by c-store service and export adapter. Main difference - it can receive (error) response while transmitting data stream.
- updated to java 11. Initially because I tried to use native new java http client. Later left as is, since there seem to be no drawbacks.

Baked both pull requests into one because export adapter got broken pipe errors while trying to use old stowrs implementation (exactly same as #5), but when I tried to replace client with Jetty to actually see the error, it just worked. Doubt it's worth looking in too deeply, since that's what I was going to do next anyway.

Note: Jetty client implements only stow-rs (POST with binary data). I don't see particularly good reason to over-complicate GET requests with use of low level api.

### DIFF
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
@@ -29,6 +29,7 @@ import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.FuturePromise;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.json.JSONArray;
+import org.json.JSONException;
 import org.json.JSONObject;
 
 public class DicomWebClientJetty implements IDicomWebClient {
@@ -79,6 +80,8 @@ public class DicomWebClientJetty implements IDicomWebClient {
       }
       requestFields.add(HttpHeader.CONTENT_TYPE,
           "application/dicom");
+      requestFields.add(HttpHeader.ACCEPT,
+          "application/dicom+json");
       MetaData.Request request = new MetaData.Request("POST", uri, HttpVersion.HTTP_2,
           requestFields);
       HeadersFrame headersFrame = new HeadersFrame(request, null, false);
@@ -132,11 +135,15 @@ public class DicomWebClientJetty implements IDicomWebClient {
 
       int httpStatus = responseCodeFuture.get();
       if (httpStatus != HttpStatus.OK_200) {
-        JSONObject responseJson = new JSONObject(resultBuilder.toString());
-        throw new DicomWebException("Http_" + responseCodeFuture.get()
-            + ", " + responseJson.getJSONObject("error").getString("status")
-            + ", " + responseJson.getJSONObject("error").getString("message"),
-            httpStatus, Status.ProcessingFailure);
+        try {
+          JSONObject responseJson = new JSONObject(resultBuilder.toString());
+          throw new DicomWebException("Http_" + httpStatus
+              + ", " + responseJson.getJSONObject("error").getString("status")
+              + ", " + responseJson.getJSONObject("error").getString("message"),
+              httpStatus, Status.ProcessingFailure);
+        } catch (JSONException e) {
+          throw new DicomWebException("Http_" + httpStatus, httpStatus, Status.ProcessingFailure);
+        }
       }
     } catch (Exception e) {
       if (e instanceof DicomWebException) {
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
@@ -43,7 +43,7 @@ public class DicomWebClientJetty implements IDicomWebClient {
       OAuth2Credentials credentials,
       String serviceUrlPrefix) {
     this.credentials = credentials;
-    this.serviceUrlPrefix = Util.trim(serviceUrlPrefix);
+    this.serviceUrlPrefix = StringUtil.trim(serviceUrlPrefix);
   }
 
   @Override
@@ -64,7 +64,7 @@ public class DicomWebClientJetty implements IDicomWebClient {
       client.addBean(sslContextFactory);
       client.start();
 
-      HttpURI uri = new HttpURI(serviceUrlPrefix + "/" + Util.trim(path));
+      HttpURI uri = new HttpURI(serviceUrlPrefix + "/" + StringUtil.trim(path));
 
       FuturePromise<Session> sessionPromise = new FuturePromise<>();
       client.connect(sslContextFactory, new InetSocketAddress(uri.getHost(), CONNECT_PORT),
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
@@ -1,0 +1,215 @@
+package com.google.cloud.healthcare;
+
+import com.github.danieln.multipart.MultipartInput;
+import com.google.auth.oauth2.OAuth2Credentials;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.InetSocketAddress;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import org.dcm4che3.net.Status;
+import org.eclipse.jetty.http.HttpFields;
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.http.HttpURI;
+import org.eclipse.jetty.http.HttpVersion;
+import org.eclipse.jetty.http.MetaData;
+import org.eclipse.jetty.http.MetaData.Response;
+import org.eclipse.jetty.http2.api.Session;
+import org.eclipse.jetty.http2.api.Stream;
+import org.eclipse.jetty.http2.api.server.ServerSessionListener;
+import org.eclipse.jetty.http2.client.HTTP2Client;
+import org.eclipse.jetty.http2.frames.DataFrame;
+import org.eclipse.jetty.http2.frames.HeadersFrame;
+import org.eclipse.jetty.http2.frames.ResetFrame;
+import org.eclipse.jetty.util.Callback;
+import org.eclipse.jetty.util.FuturePromise;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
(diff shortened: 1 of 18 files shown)


dicom_util/src/main/java/com/google/cloud/healthcare/Util.java
```diff
@@ -0,0 +1,10 @@
+package com.google.cloud.healthcare;
```
danielbeaudreau: I'm not sure it's worth introducing a new class for just this one method. I think duplicating it is fine given it's just one line.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -0,0 +1,215 @@
...
+      }
+    } catch (Exception e) {
+      if (e instanceof DicomWebException) {
+        throw (DicomWebException) e;
+      }
+      throw new DicomWebException(e);
+    }
+  }
+
+  private static class DataStream {
```
danielbeaudreau: The Jetty implementation doesn't provide any utilities for streaming data? I would have thought there would be something we could use instead of having to write these lower level classes ourselves

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
```diff
@@ -120,8 +120,8 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
 
     Subscriber subscriber = null;
     try {
-      SubscriptionName subscriptionName =
-          SubscriptionName.of(Flags.projectId, Flags.subscriptionId);
+      ProjectSubscriptionName subscriptionName =
```
danielbeaudreau: Why the change of class here? Java 11 compatibility?

red1408:
> Any chance you could squash this PR so it's fewer commits, rather than 91?

Right, force-pushed branch with only one commit. I'm not sure why it ends up showing 91, when created directly from master (at least changes are actually all the same).

dicom_util/src/main/java/com/google/cloud/healthcare/Util.java
red1408: I just really dislike code duplication. There could be more clients later too (well, there were more, during testing/working it out phase).

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
red1408: No, there doesn't seem to be a class for this.

There is whole high level Jetty client though (HttpClient class as opposed to low level Http2Client used by me). And it almost works, but ends up violating flow control during send more often than not (sending more data than allowed by WINDOW_UPDATE). I'm not sure how to fix this in high level client (If I did, I wouldn't need the low level implementation).

While current low level implementation just seems to work (I send next frame after callback succeeds, so no flow control errors there).

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
red1408: More specifically changes in 'com.google.cloud:google-cloud-pubsub:1.89.0' (I couldn't keep old version due to dependencies conflict).
From what I understand it's exactly same thing, but renamed.

dicom_util/src/main/java/com/google/cloud/healthcare/Util.java
danielbeaudreau: Ok, can we rename this class to StringUtil instead?

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -0,0 +1,222 @@
...
+          requestFields);
+      HeadersFrame headersFrame = new HeadersFrame(request, null, false);
+
+      // Prepare the listener to receive the HTTP response frames.
+      final StringBuilder resultBuilder = new StringBuilder();
+      final CompletableFuture<Integer> responseCodeFuture = new CompletableFuture<>();
+      final CompletableFuture<Boolean> doneFuture = new CompletableFuture<>();
+      Stream.Listener responseListener = new Stream.Listener.Adapter() {
+        @Override
+        public void onReset(Stream stream, ResetFrame frame) {
```
danielbeaudreau: When does onReset get called?

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -0,0 +1,222 @@
...
+      SslContextFactory sslContextFactory = new SslContextFactory.Client();
+      client.addBean(sslContextFactory);
+      client.start();
+
+      HttpURI uri = new HttpURI(serviceUrlPrefix + "/" + Util.trim(path));
+
+      FuturePromise<Session> sessionPromise = new FuturePromise<>();
+      client.connect(sslContextFactory, new InetSocketAddress(uri.getHost(), CONNECT_PORT),
+          new ServerSessionListener.Adapter(), sessionPromise);
+      Session session = sessionPromise.get(5, TimeUnit.SECONDS);
```
danielbeaudreau: 5 seconds might be too low for very large files. can we increase this limit to like 10 minutes?

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
danielbeaudreau: Ok, can you please add a comment explaining why we need the low level method? (flow control error) Thanks

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
danielbeaudreau: I see, could you please test manually that the export adapter still works after this change?
We should probably have an integration test for the export adapter if we are making changes to it, but it's not really a high priority task right now.

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
```diff
@@ -79,22 +83,18 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
...
-      } else {
-        requestFactory = createHttpRequestFactory(null);
-      }
-      DicomWebClient exportDicomWebClient =
-          new DicomWebClient(requestFactory, flags.peerDicomwebAddr);
+      boolean isLegacyAdress = flags.peerDicomwebAddress.isEmpty();
+      String peerDicomwebAddress =
+          isLegacyAdress ? flags.peerDicomwebAddr : flags.peerDicomwebAddress;
+      String peerDicomwebStowpath = isLegacyAdress ? flags.peerDicomwebStowPath : "studies";
+      IDicomWebClient exportDicomWebClient =
```
danielbeaudreau: Would any post request succeed without this flag set to true? How would we get any credentials for the request?

danielbeaudreau:
What do you think about moving the jetty stow implementation into the default dicomwebclient? Because cmove also needs stow, however it will be using the non-jetty implementation. Ideally all callers would use the same jetty implementation.

dicom_util/src/main/java/com/google/cloud/healthcare/Util.java
red1408: Ok

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
red1408: I already tested it manually, that's the point where I had to replace old stowrs implementation with jetty one.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
red1408: It's timeout to establish the session, it doesn't affect data transmission.

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
red1408: I assumed sending request without auth header was the intention here, but after looking deeper it doesn't seem to be right.

Though I'm also not sure under which conditions this flag was useful.
credentials are obtained as:

```
GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
if (!flags.oauthScopes.isEmpty()) {
      credentials = credentials.createScoped(Arrays.asList(flags.oauthScopes.split(",")));
}
```

The flag was used  like this:

```
// DicomWeb client for sink of DICOM.
// Use application default credentials for HTTP requests of DICOM sink, if specified by flag.
      HttpRequestFactory requestFactory = null;
      if (flags.useGcpApplicationDefaultCredentials) {
        requestFactory = createHttpRequestFactory(credentials);
      } else {
        requestFactory = createHttpRequestFactory(null);
      }
      DicomWebClient exportDicomWebClient =
          new DicomWebClient(requestFactory, flags.peerDicomwebAddr);
```

... So, assuming comments here are correct, either way it's application default credentials, with scopes being the only difference.

Can I just remove or ignore this flag?

red1408:
> What do you think about moving the jetty stow implementation into the default dicomwebclient? Because cmove also needs stow, however it will be using the non-jetty implementation. Ideally all callers would use the same jetty implementation.

C-move instructs receiver to execute c-store against c-move target. There is no stow involved.
I could make a composite IDicomwebclient implementation when it becomes necessary, but at current point there is no service using all 3 dicomweb methods.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
red1408: When error response comes during post body transmission. Frame sequence for error is Header  -> Data (not end stream)  -> Reset.
For normal response it's Header  -> Data (end stream).

export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java
danielbeaudreau: From the comment on Flags.java: 
"If true, uses GCP Application Credentials to when sending HTTP requests to peer. This is useful if the peer DICOMWeb endpoint is running in GCP."
So I realized that this is for non-GCP dicomWeb backends. So it's fine to leave as-is.

danielbeaudreau:
Looks good to me. Can you please add a follow up PR to delete the old STOW implementation in the non-jetty DicomWebClient, and instead throw an unimplemented error?

Issue #22, opened 2019-07-23
Support DeID'ing files as they go through the adaptor


dzelemba:
Using https://github.com/GoogleCloudPlatform/healthcare-deid/tree/master/offline_tools/redactor

pavertomato:
do we have any kind of maven repo where jar of this tool is available? or it have to be compiled with adapter?

dzelemba:
I think the team was working on making it available through maven, let me check.

danielbeaudreau:
Thanks Egor, can you please create a PR to update the readme as well?

red1408:
Sure, will do it tomorrow.

PR #43, opened 2019-10-15
Support DeID'ing files as they go through the adaptor #22

red1408:
Currently integration test uses deid-redactor from my fork. I also created pull request for deid project:
https://github.com/GoogleCloudPlatform/healthcare-deid/pull/12


cloudbuild.yaml
```diff
@@ -1,5 +1,35 @@
...
+- name: ${_GRADLE_IMAGE}
+  volumes:
+  - name: 'vol1'
+    path: '/root/.m2'
   args:
   - gradle
   - build
+
+substitutions:
+  _GRADLE_IMAGE: 'gradle:5.6-jdk11'
```
danielbeaudreau: We shouldn't be forcing users to clone this repo in order to build the adapter.
While we wait for the maven package, is it possible for us to use jitpack.io to package the dependency here?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -64,6 +70,9 @@ protected void store(
       String transferSyntax = presentationContext.getTransferSyntax();
       String remoteAeTitle = association.getCallingAET();
 
+      log.info("Received C-STORE for association {}, SOP class {}, TS {}, remote AE {}",
```
danielbeaudreau: Let's spell out transfer syntax since the logs are user visible. Also should we have logging like this for the other services like CFIND, MOVE etc?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -72,12 +81,42 @@ protected void store(
...
+          try {
+            dicomWebClient.stowRs(path, pipedIn);
+            futureThrowable.complete(null);
+          } catch (Throwable e) {
+            futureThrowable.complete(e);
+            try {
+              pipedIn.close();
+            } catch (IOException e1) {
+              log.error("Failed to close pipedIn", e);
+            }
```
danielbeaudreau: Should this be in a finally {} block? Do we always want to close pipedIn?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -72,12 +81,42 @@ protected void store(
...
+          }
+        });
+
+        try {
+          redactor.redact(inBuffer, pipedOut);
+        } catch (Exception e){
+          // not sure if this isn't race condition
+          if(!futureThrowable.isDone()) {
+            throw e;
+          }
```
danielbeaudreau: If futureThrowable is done but the result is null, then we're not throwing an error here when we should be. Can we just throw e here, and if futureThrowable has an exception we will throw it in the next block?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
```diff
@@ -92,6 +92,32 @@
...
+  @Parameter(
+      names = {"--tag_edit_keep"},
+      description = "Tags to keep during C-STORE upload, comma separated. Only one of 'tag_edit' group flags may be present"
+  )
+  String tagsToKeep = "";
+
+  @Parameter(
+      names = {"--tag_edit_profile"},
+      description = "Filter tags by predefined profile during C-STORE upload. Only one of 'tag_edit' group flags may be present. "
+      + "Values: TAG_FILTER_PROFILE_UNSPECIFIED, CHC_BASIC"
```
danielbeaudreau: Is TAG_FILTER_PROFILE_UNSPECIFIED an actual possible value a user would specify or just a default?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -107,4 +119,36 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
...
     device.bindConnections();
   }
+
+  private static IDicomRedactor configureRedactor(Flags flags) throws IOException{
+    IDicomRedactor redactor = null;
+    int tagEditFlags = (flags.tagsToRemove.isEmpty() ? 0 : 1) +
+        (flags.tagsToKeep.isEmpty() ? 0 : 1) +
+        (flags.tagsProfile.isEmpty() ? 0 : 1);
+    if (tagEditFlags > 1) {
+      throw new IOException("Only one of 'tag_edit' flags may be used");
```
danielbeaudreau: IOException doesn't seem to fit this error case, maybe java.lang.IllegalArgumentException?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
```diff
@@ -76,7 +76,7 @@ private int createDicomServer(boolean connectError, int responseCode) throws Exc
     IDicomWebClient dicomWebClient =
```
danielbeaudreau: Can we add a couple unit tests for redaction?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
```diff
@@ -92,6 +92,32 @@
   )
   String monitoringProjectId = "";
 
+  @Parameter(
+      names = {"--tag_edit_remove"},
```
danielbeaudreau: Nitpick: can we rename these to:
--redact_remove_list
--redact_keep_list
--redact_filter_profile?
I think using the full name gives more details, especially for the filter profile. Using redact as the prefix helps the user understand that these are for the redact library.

red1408:
After checking more in-depth, I found a problem with my approach:
DicomInputStream.readDataset either fully reads a Sequence or Tag value, or not at all. Which means that large sequences and binary tags (like PixelData, which typically takes most of dcm instance size) will be read in one pass anyway.
This doesn't result in failure to process such instances, but point of trying to process them in chunks is mostly lost.
Should I scale down pull request to redactor to just maven artifact related part?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: This part is quite literal, wrong value will fail adapter startup like this:
java.lang.IllegalArgumentException: No enum constant com.google.cloud.healthcare.deid.redactor.protos.DicomConfigProtos.DicomConfig.TagFilterProfile.NO_SUCH_VALUE

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
red1408: Redactor is unit tested in it's own project, so I don't think it would be a good idea to duplicate it in adapter.
I could add it to integration test though.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: It may be actually best to remove this line, since it's somewhat legacy. Dcm4che already logs it all, in a good format.

(last message is from this line, rest is dcm4che standard)

```
DEBUG org.dcm4che3.net.Association  - IMPORTADAPTER<-STORESCU(1): enter state: Sta6 - Association established and ready for data transfer
INFO  org.dcm4che3.net.Dimse  - IMPORTADAPTER<-STORESCU(1) >> 1:C-STORE-RQ[pcid=5, prior=0
  cuid=1.2.840.10008.5.1.4.1.1.1.2 - Digital Mammography X-Ray Image Storage - For Presentation
  iuid=1.3.6.1.4.1.5962.1.178.1174395404.11033.0 - ?
  tsuid=1.2.840.10008.1.2.1 - Explicit VR Little Endian
DEBUG org.dcm4che3.net.Dimse  - Command:
(0000,0002) UI [1.2.840.10008.5.1.4.1.1.1.2] AffectedSOPClassUID
(0000,0100) US [1] CommandField
(0000,0110) US [1] MessageID
(0000,0700) US [0] Priority
(0000,0800) US [0] CommandDataSetType
(0000,1000) UI [1.3.6.1.4.1.5962.1.178.1174395404.11033.0] AffectedSOPInstance

INFO  com.google.cloud.healthcare.imaging.dicomadapter.CStoreService  - Received C-STORE for association IMPORTADAPTER<-STORESCU(1), SOP class 1.2.840.10008.5.1.4.1.1.1.2, TS 1.2.840.10008.1.2.1, remote AE STORESCU
```

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Ok

cloudbuild.yaml
red1408: Working on it, couldn't get it to work right away.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: redact has both input and output in try-with-resources, so ultimately closing streams is not in question. Closing it from here is done to intentionally fail redact (as opposed to having it wait indefinitely after stowRs already failed).

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Problem is, exception in redact causes exception in stowRs and I need vice-versa to interrupt redact of failed upload too(so I close pipedIn) . Intention here is to throw up only original cause and not the consequence.

cloudbuild.yaml
danielbeaudreau: Ok, feel free to post any issues you may have here.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: I'm good with removing this line.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
danielbeaudreau: Sure, but I don't think users will ever want to specify TAG_FILTER_PROFILE_UNSPECIFIED here. Right?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
danielbeaudreau: Makes sense, let's add it to the integration test.

danielbeaudreau:
I see, in that case I would suggest focusing on getting the maven repository issue resolved first, and maybe we can try to get streaming working afterwards.
What is the impact of loading the entire instance in memory in the redactor? I guess we will have a risk of OOMExceptions for large instances.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: I see, can you please add a small comment explaining this?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: If an exception in redact causes a stow exception, then potentially futureThrowable.isDone will already be true here and we will throw the stow-RS exception incorrectly, right?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Ok, I'll remove the mention here (it does same thing as CHC_BASIC anyway).

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Yes, but I don't see how I could bypass the potential problem. If I added synchronized blocks, it would become matter of which is entered first, for same result.

I could change redact signature to take completable future argument, to be able to complete it inside, thus guaranteeing order. But this would put stuff that is rather irrelevant to redactor library function into it.

danielbeaudreau: I'm not sure if a redact exception would cause a stow-RS exception, stow-RS might just hang waiting for input? Is it possible to test that case?

I'm also not sure if pipedIn.close() will necessarily fail the stow-RS call. What if it closes the file at a point where it is almost fully read, so the server can parse it, and accepts it?

An alternative approach I found was to use two threads and an ExecutorCompletionService to get the first result. See an example here: https://stackoverflow.com/a/29116763  it might make this error handling simple if the above is an issue. What do you think?

cloudbuild.yaml
danielbeaudreau: What were the issues you ran into with jitpack.io? Anything we can do to unblock this?

red1408: It does expect single project in repository, with build file at top level, not collection of projects. It is possible to write custom script in jitpack.yml at top level, but there is no documentation for this (https://jitpack.io/docs/BUILDING/#custom-commands is all there is on topic, and neither looking through first few pages of jitpack issues on github helped). I got it to install the project to local maven, but then jitpack fails to find it ('ERROR: No build artifacts found' after gradle finished build).

I could either try to extract redactor-deid-lib into self-contained project, or maybe go entirely different way, using it as git submodule from adapter.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: As far as I've tested, both interrupted each-other reliably. I'm not sure how server would handle the prematurely terminated data stream though.

Seems like this approach should work, thanks for finding it. I'd technically use 3 threads - original to pipe PDVInputStream (which can't be read on any other, dcm4che limitation), 2nd to redact and 3rd to send.

cloudbuild.yaml
danielbeaudreau: I did some searching and it couldn't find much either. If you can't get the custom build script to work, we might need to wait for a maven repo for the redactor.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: Yeah, I'm not sure if the executor service strongly guarantees order, but it should make killing the other task easy. If it turns out that we sometimes get the wrong exception, I guess it's OK as long as it's rare.

When we kill the stow thread, I wonder if the server sees this as a cancellation (similar to a CTRL+C in curl) and returns an error. Were you able to see the HTTP response in this error scenario during testing?

Do we need any clean up logic here to avoid leaking threads?

red1408: After checking for leaks with VisualVM - yes, you are right. Any situation where either thread failed leaked the other, fixed by enclosing all pipe usages in try-with-resources.

If redactor fails while reading or actually redacting, stowRs thread gets 409 with the following details:
```
{"00081198":{"Value":[{"00091097":{"Value":["empty DICOM instance found"],"vr":"LT"},"00090010":{"Value":["GOOGLE"],"vr":"LO"},"00081197":{"Value":[272],"vr":"US"}}],"vr":"SQ"}}
```

I suspect if it somehow failed mid-write, that could be a problem. But doesn't seem likely, at least with current read-all, process-all, write-all sequential implementation.

cloudbuild.yaml
red1408: Extracted redactor to separate repo, using it from jitpack.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
red1408: Will finish it on monday, probably.

import/build.gradle
```diff
@@ -49,6 +49,7 @@ repositories {
     mavenCentral()
     maven { url 'http://www.dcm4che.org/maven2/' }
     maven { url 'https://jitpack.io' }
+    mavenLocal()
```
danielbeaudreau: Is mavenLocal still necessary?

import/build.gradle
```diff
@@ -61,6 +62,8 @@ dependencies {
     compile project(":dicom_util")
     compile project(":util")
 
+    compile "com.github.red1408:deid-redactor:233908ca"
```
danielbeaudreau: Dusan said that he preferred us embedding a precompiled .jar from the deid redactor project rather than creating a clone repo. Are you ok with that change? I.e. building a .jar file and hosting it in this repo, and then having gradle compile it here.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -33,6 +33,9 @@
 
 - name: ${_GRADLE_IMAGE}
   id: adapter-build
+  volumes:
+  - name: 'vol1'
+    path: '/root/.m2'
```
danielbeaudreau: Is this code still necessary?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: Thanks! So even if the stow-RS command fails at the end of the pipeline, the earlier threads will be interrupted still because they will be writing to a closed stream?

red1408: Yes, though I suppose redactor will do some unnecessary work in between (since it still reads input, processes it and only then tries to write and fails).

import/build.gradle
red1408: You are right, removed it.

integration_test/cloudbuild-integration-test.yaml
red1408: Removed it

cloudbuild.yaml
red1408: Update integration test use redactor, this involves few extra changes to work around UID replacement that redactor does by default (not configurable). Also organized integration test files into scripts and data subfolders.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
red1408: Updated integration test to use redactor, this involves few extra changes to work around UID replacement that redactor does by default (not configurable). Also organized integration test files into scripts and data subfolders.

import/build.gradle
red1408: Ok, changed to pre-built redactor jar.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
danielbeaudreau: Yeah I think that's fine.

Issue #42, opened 2019-10-08
Support QIDO-RS pagination when handling C-FIND requests


red1408:
The default number of results is 100 for studies/series and 1,000 for instances.

red1408:
Ok, added to C-FIND section of conformance statement.

danielbeaudreau:
We want to convey some warning to the user when there are 50,000 C-FIND results indicating that there may be more results. But, C-FIND doesn't support the "warning" status. Egor can you please investigate how other C-FIND implementations handle this scenario to determine what the standard is? Thanks.

red1408:
Right, there is no standard "warning" status for C-FIND.
As for what others do: I've looked through first several pages worth of DICOM Conformance statements for various stuff on google. C-FIND result limit was mentioned a few times, but only in SCUs (not a problem at SCU side, just cancel on hitting limit). So this provided no solution as well.

As for more abstract ideas, I suppose I could append a fake result with potentially displayable fields stuffed with "Hit result limit" in some way, hoping that user will notice the unusual result. But this is quite dubious - a user that attentive would already notice that 5000/50000 is too round number anyway. And depending on how client sorts results presented to user, the fake result might be hidden somewhere in the middle of result list.

red1408:
@danielbeaudreau 
As I've noted above, a good way to do this doesn't seem to exist.

danielbeaudreau:
Since we have this documented in our conformance statement, and there's no way to include a warning in a c-find response, I think we can close this for now.

PR #44, opened 2019-10-28
Set maximum limits for qido-rs requests #42

red1408:
As discussed with Dusan, just increased request limits to maximum (50000/5000 instead of default 1000/100)


util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -114,13 +117,13 @@ public static String attributesToQidoPath(Attributes attrs, String... includeFie
     if (nonEmptyKeys.contains(Tag.QueryRetrieveLevel)) {
       switch (attrs.getString(Tag.QueryRetrieveLevel)) {
         case "STUDY":
-          qidoPath.append("studies");
+          qidoPath.append("studies?limit=" + STUDIES_SERIES_LIMIT + "&");
```
danielbeaudreau: I think we should have an integration test step for studies and series C FIND requests since they're a different code path, can you please add this in a follow up PR?

red1408: Ok, added.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -143,17 +155,46 @@
...
   dir: "dcm4che/dcm4che-tool/dcm4che-tool-movescu"
   waitFor:
-  - run-commitment-scu
+  - run-store-scu
   - wait-for-storescp
 
+# Stgcmtscu  waits for n-event-report after n-action request returns success,
+# if it doesn't get one within cloud build step timeout of 30 seconds, step fails.
+# Run it after other tools to avoid timeout due to parallel execution.
+- name: ${_MAVEN_IMAGE}
```
danielbeaudreau: This is taking more than 30s to complete when there are other requests?

red1408: Each cloud build step adds about 5s delay just for container initialization/teardown. So it's not adapter itself that takes 30s to process few minimalistic queries, commitment step failed because it didn't initialize in time.

Issue #50, opened 2019-11-11
Error Comment (0000,0902) wrongly passed in data attributes instead of cmd for C-FIND and C-MOVE


red1408:

danielbeaudreau:
Good catch

PR #51, opened 2019-11-12
[#50] Error Comment (0000,0902) was wrongly passed in data attributes…

red1408:
… instead of cmd for C-FIND and C-MOVE

### DIFF
--- a/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
+++ b/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
@@ -135,15 +135,13 @@ public class CFindService extends BasicCFindSCP {
         MonitoringService.addEvent(Event.CFIND_CANCEL);
         as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.Cancel));
       } catch (IDicomWebClient.DicomWebException e) {
-        log.error("CFind qido-rs error", e);
+        log.error("CFind Qido-rs error", e);
         MonitoringService.addEvent(Event.CFIND_QIDORS_ERROR);
-        as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, e.getStatus()), e.getAttributes());
+        sendErrorResponse(e.getStatus(), e.getMessage());
       } catch (Throwable e) {
         log.error("Failure processing CFind", e);
         MonitoringService.addEvent(Event.CFIND_ERROR);
-        Attributes attrs = new Attributes();
-        attrs.setString(Tag.ErrorComment, VR.LO, e.getMessage());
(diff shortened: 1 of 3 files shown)


import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
```diff
@@ -231,20 +231,16 @@ private void sendErrorResponse(int status, String message, List<String> failedIn
...
-        attributes.setString(Tag.ErrorComment, VR.LO, message);
+        cmdAttr.setString(Tag.ErrorComment, VR.LO, message);
       }
       if (failedInstanceUids != null && failedInstanceUids.size() > 0) {
-        attributes.setString(Tag.FailedSOPInstanceUIDList, VR.UI,
+        dataAttr.setString(Tag.FailedSOPInstanceUIDList, VR.UI,
             failedInstanceUids.toArray(new String[]{}));
       }
-      as.tryWriteDimseRSP(pc, Commands.mkCMoveRSP(cmd, status), attributes);
+      as.tryWriteDimseRSP(pc, cmdAttr, dataAttr);
```
danielbeaudreau: Due to the way dcm4che is implemented, tryWriteDimseRSP calls into writeDimseRSP which when the dataset parameter is not null, sets CommandDataSetType to "with dataset" type instead of "no dataset" type. However there is no dataset. So we should avoid passing a dataset with 0 objects into tryWriteDimseRSP

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindService.java
```diff
@@ -135,15 +135,13 @@ public void run() {
         MonitoringService.addEvent(Event.CFIND_CANCEL);
         as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, Status.Cancel));
       } catch (IDicomWebClient.DicomWebException e) {
-        log.error("CFind qido-rs error", e);
+        log.error("CFind Qido-rs error", e);
         MonitoringService.addEvent(Event.CFIND_QIDORS_ERROR);
-        as.tryWriteDimseRSP(pc, Commands.mkCFindRSP(cmd, e.getStatus()), e.getAttributes());
+        sendErrorResponse(e.getStatus(), e.getMessage());
```
danielbeaudreau: We can remove the getAttributes() method on DicomWebException since there are no other callers.

red1408: Ok

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CMoveService.java
red1408: Nice find! Fixed.

Issue #54, opened 2019-12-02
Issue with query/retrieve status


cdoshi:
Hi,

I am executing the following command
`movescu -v -k 0008,0052=STUDY -k AccessionNumber=xxx -aet IMPORTADAPTER -aec SCP xxx.xxx.xx.xx 104
` from the terminal

If I run it once, it throws the following error `Move response with error status (Refused: OutOfResourcesSubOperations)` but if I check the data store the images have been moved and are available. If I run the same command again, it runs successfully. 

Now if I delete the study from the store and run the command again, I again get the OutOfResources error. I see the same behavior with pynetdicom send_c_move as well.

Does anyone know why this is happening?

red1408:
Hi, need a few more details.

Is error status 0xA700 (OutOfResources) or 0xA702 (UnableToPerformSubOperations) ? "OutOfResourcesSubOperations" falls in between dcm4che terminology I'm more familiar with.

What do you run the adapter against (healthcare api or something else)?

cdoshi:
Hi @red1408, I am getting a 0xA702 error. I am currently using the terminal to perform cmove between the PACS and the adapter via the load balancer.

red1408:
0xA702 happens when all C-STORE sub operations within C-MOVE fail, could you check  adapter's log output for error messages starting with "Failed CStore within CMove" ?

cdoshi:
I am trying to enable stack driver monitoring but I am unable to do so. Here is my dicom_adapter.yaml file. Any ideas?

`apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: xxx-dicom-adapter
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: xxx-dicom-adapter
    spec:
      containers:
        - name: xxx-import-adapter
          image: gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-import:0.1.12
          env:
          - name: ENV_POD_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
          - name: ENV_POD_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
          - name: ENV_CONTAINER_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          ports:
            - containerPort: xxxx
              protocol: TCP
              name: "port"
          command:
            - "/import/bin/import"
[...]

PR #55, opened 2019-12-04
Log full STOW-RS response

red1408:
Related to #54 .
Can't pass full response to DIMSE client (ErrorComment tag has very short length), but at least it should be present in log.


dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -137,6 +141,7 @@ public void onData(Stream stream, DataFrame frame, Callback callback) {
       if (httpStatus != HttpStatus.OK_200) {
         try {
           JSONObject responseJson = new JSONObject(resultBuilder.toString());
+          log.info("STOW-RS response:\n" + responseJson.toString(2));
           throw new DicomWebException("Http_" + httpStatus
```
danielbeaudreau: Can we change this to log.error?

red1408: Ok

Issue #47, opened 2019-11-05
Support routing to multiple DICOM stores


dzelemba:

mukhlin:
Hi @dzelemba 

We have the following questions:

1. Is this issue related to export or import or both (and which services)?
1. How do you want to configure list of stores?
1. What is the criteria to choose particular store?
1. Are you expect change in Adapter or it can be implemented as separate application (for example, NGINX based proxy)?

mukhlin:
Hi @danielbeaudreau,

Can you please have a look at my comment above as well?

danielbeaudreau:
Hi, I'll get back to you on this soon.

dzelemba:
More detailed requirements are available in https://docs.google.com/document/d/1pWzEUkORVm5kpNWoa9xQsGLOPleZ3MfXA91GHPzy1R8/edit

PR #57, opened 2020-02-06
[#47] Support routing to multiple DICOM stores 

red1408:
As discussed with Dusan, implementation differs a bit from initial proposal https://docs.google.com/document/d/1pWzEUkORVm5kpNWoa9xQsGLOPleZ3MfXA91GHPzy1R8 .
I kept --dicomweb_address with --destination_config_path acting as set of overrides, rather than mutually exclusive flag. Which means that services other than C-STORE keep operating exactly as before.

I also changed environment variable to look for from "DESTINATION_CONFIG_JSON" to "ENV_DESTINATION_CONFIG_JSON" for consistency (easy to change to original value, if needed).  

 These changes are reflected in Readme.

Idea is that I read ahead the tags before PixelData, decide destination based on them, then rewind the buffered stream to pass it to redactor/sender. This is possible because tags in Dicom file have to be ordered by tag value and PixelData seems to have highest non-reserved value.


util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/DestinationFilter.java
```diff
@@ -0,0 +1,81 @@
...
+    private final Attributes filterAttrs;
+
+    public DestinationFilter(String filterString) {
+        String aeTitle = null;
+        this.filterAttrs = new Attributes();
+
+        if(filterString != null && filterString.length() > 0) {
+            String[] params = filterString.split("&");
+            for (String param : params) {
+                String[] keyValue = param.split("=");
```
danielbeaudreau: Let's throw an illegal argument if len(keyValue) != 2

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -291,6 +336,7 @@
   - "--quiet"
```
danielbeaudreau: Do you need to delete the second dcm store as well?

README.md
```diff
@@ -73,6 +73,53 @@ The AET dictionary JSON can also be specified directly via the "--aet_dictionary
 
 For the list of command line flags, see [here](import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java)
 
+## Routing to multiple DICOM stores
```
danielbeaudreau: Since the readme is getting a bit long, and this feature is likely only for advanced users, let's move this section to the wiki and add a link here instead. Something like:
"The Import Adapter supports routing to multiple DICOM stores based on tags in the incoming instance and the client application entity title. Please see \[this page\](link_to_wiki) for more information."

README.md
```diff
@@ -73,6 +73,53 @@ The AET dictionary JSON can also be specified directly via the "--aet_dictionary
...
+
+Import adapter C-STORE service supports routing to multiple DICOM stores based on tags in incoming instance and client AET.
+Filter is in the format fo a QidoRS query, with addition of AETitle virtual tag for matching calling client AET.
```
danielbeaudreau: "fo" should be "of"

README.md
```diff
@@ -73,6 +73,53 @@ The AET dictionary JSON can also be specified directly via the "--aet_dictionary
 
 For the list of command line flags, see [here](import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java)
 
+## Routing to multiple DICOM stores
+
+Import adapter C-STORE service supports routing to multiple DICOM stores based on tags in incoming instance and client AET.
```
danielbeaudreau: wording nit pick: "tags in the incoming instance and the client AE title."

README.md
danielbeaudreau: I think calling it a virtual tag might confuse some readers. How about:
"Filter is in the format of a QidoRS query, with the additional option of specifying the AETitle for matching against the calling client AET."

danielbeaudreau:
We should also add a warning that matching tags with a hex value greater than pixeldata (7FE0,0010) is not supported

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -61,14 +61,16 @@ public JSONArray qidoRs(String path) throws DicomWebException {
   }
 
   @Override
-  public void stowRs(String path, InputStream in) throws DicomWebException {
+  public void stowRs(InputStream in) throws DicomWebException {
     try {
+      log.debug("STOW-RS to: " + stowPath);
```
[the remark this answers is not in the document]
red1408: I'd like to leave this one. There is no other point when adapter would log which destination it chose.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -75,15 +79,29 @@ protected void store(
       validateParam(sopClassUID, "AffectedSOPClassUID");
       validateParam(sopInstanceUID, "AffectedSOPInstanceUID");
 
-      CountingInputStream countingStream = new CountingInputStream(inDicomStream);
-      InputStream inBuffer =
+      CountingInputStream countingStream;
+      IDicomWebClient destinationClient;
+      if(destinationMap != null && destinationMap.size() > 0){
+        DicomInputStream inDicomStream  = new DicomInputStream(inPdvStream);
+        inDicomStream.mark(Integer.MAX_VALUE); // do or die (OOM)
```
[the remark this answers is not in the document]
red1408: In theory this could get out of memory, since mark parameter sets size limit for BufferedInputStream inside the DicomInputStream.
But it's unlikely unless instance has unreasonable amount of OB (other binary) tags or sequences.
I think it's safe to remove this comment.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/DestinationFilter.java
red1408: Ok

integration_test/cloudbuild-integration-test.yaml
red1408: Well, it won't break the test, since deleting pre-existing stores is part of setup.
But sure, makes sense to delete it here.

README.md

README.md

README.md
red1408: Ok, extracted to wiki.

red1408:
All done, including the warning about tag hex value.

Issue #56, opened 2020-01-16
Support in-transit compression for C-STORE -> StowRS


dzelemba:
Feature request to support a command line parameter that allows a user to configure the adapter such that all instances that pass through C-STORE are compressed before being sent to StowRS.

--store_compress_to_transfer_syntax=<TransferSyntaxUID>

Care should be taken in implementation to minimize the memory overhead required.

To avoid using local disk or having to store large instance fully in memory, no retry behavior will be added in the case of compression failures (which should be very rare).

danielbeaudreau:
To add some more context here:
- To start we only need to support one lossless compression syntax. Dusan and I thought that JPEG2000 will be the best fit, because the healthcare API supports decompressing it easily, and there exists a native Java library for compressing jpeg 2000.
- We don't want to use any C/C++ libraries due to security concerns. We would like to keep the compression libraries java only.
- Here is an open source JPEG 2000 library that you can use: https://jai-imageio.github.io/jai-imageio-jpeg2000/javadoc/com/sun/media/imageioimpl/plugins/jpeg2000/J2KImageReader.html
- You can configure the dcm4che ImageWriterFactory to use this plugin by modifying the properties file in this directory: https://github.com/dcm4che/dcm4che/blob/master/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/ImageWriterFactory.java#L164
The properties file should contain the following config:

1.2.840.10008.1.2.4.90:jpeg2000:com.sun.media.imageioimpl.plugins.jpeg2000.J2KImageWriter::
1.2.840.10008.1.2.4.91:jpeg2000:com.sun.media.imageioimpl.plugins.jpeg2000.J2KImageWriter::

It might be difficult to do this with gradle. 
Please let me know if you have any concerns about the above comments.

red1408:
Overall configuring dcm4che to use the above java lib proved to be fairly easy, but while modifying integration test I run into an issue with it.

Scenario:
1) Original jpeg image is c-stored and transcoded in process to jpeg2k with java lib above, then downloaded (with curl).
2) I c-store jpeg2k image resulting from previous step, adapter is configured to transcode back to jpeg with java lib, and get the below stacktrace. Exactly same c-store command, **but using native opencv lib** passes without problems.

[...]

PR #61, opened 2020-03-12
[#56] C-STORE in-transit transcoding

red1408:
- updated dcm4che to 5.18.1 (which lead to some compatibility changes)
- using custom fork of jai-imageio-jpeg2000 to solve DICOM padding problem
- using "application/dicom; transfer-syntax=*" instead of multipart in wado requests
(because using multipart means requesting server to convert file to explicitVR, which will fail with jpg2000)

I did create a pull request for original library https://github.com/jai-imageio/jai-imageio-jpeg2000/pull/33 as discussed, but I don't think it's likely to be accepted:
- "There is NO FURTHER DEVELOPMENT in this repository; any commits here are just to keep the build working with recent versions of Maven/Java."
- It's a fix necessary for use of jai-imageio-jpeg2000 within specific dcm4che context, not a generic case.


red1408:
There are several jpeg2000 plugins inside geosolutions repo, but:
- The most straightforward one, Direct Kakadu Jpeg2000 Plugin, uses Kakadu library binaries internally. Which are proprietary. Imo, there is no point in using this over opencv implementation. In fact, I couldn't even find binaries to test with anywhere (probably inside kakadu sdk, which needs to be purchased).
- Everything else (MrSID and ECW jpeg2000 plugins) uses GDAL internally (which also uses huge amount of binaries, but at least they are available via apt-get). And for some weird reason GDAL needs URIImageInputStream while dcm4che passes only an ImageInputStream implementation. So GDAL fails to recognize it as proper image input stream.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
```diff
@@ -66,7 +66,7 @@ public void testAttributesToQidoPath_urlEncode() throws Exception {
     String result = AttributesUtil.attributesToQidoPath(attrs);
 
     assertThat(result).isEqualTo(
-        "instances?limit=50000&" + TagUtils.toHexString(Tag.PatientName) + "=%25%26%23%5E+&");
+        "instances?limit=50000&" + TagUtils.toHexString(Tag.PatientName) + "=%25%26%23%5E&");
```
danielbeaudreau: Not sure what this change means?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
```diff
@@ -131,6 +131,11 @@
   )
   String destinationConfigInline = "";
 
+  @Parameter(
+      names = {"--store_compress_to_transfer_syntax"},
```
danielbeaudreau: We might add support for gzip in the future. So can we rename the flag to "store_compression_type" and document that this means a transfer syntax (for now)?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
```diff
@@ -134,7 +134,6 @@ public void basicCFindServiceTest(IDicomWebClient serverDicomWebClient,
     // Issue CFIND
     DimseRSPAssert rspAssert = new DimseRSPAssert(association, expectedStatus);
     association.cfind(
-        UID.StudyRootQueryRetrieveInformationModelFIND,
```
danielbeaudreau: Why did this change?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -247,7 +261,7 @@
   args:
   - diff
   - 'integration_test/downloaded.dcm'
-  - 'integration_test/data/example-redacted.dcm'
+  - 'integration_test/data/example-redacted-jp2k.dcm'
```
danielbeaudreau: Can you delete the old no longer used files?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -151,11 +151,7 @@ public static String attributesToQidoPath(Attributes attrs, String... includeFie
 
       for (String value : values) {
         String encodedValue;
-        try {
```
danielbeaudreau: Why was this change made?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
```diff
@@ -308,14 +327,16 @@ private void basicCStoreServiceTest(
       int expectedDimseStatus,
       MockDestinationConfig[] destinationConfigs,
       String sopClassUID,
-      String sopInstanceUID) throws Exception {
+      String sopInstanceUID,
+      String testFile,
```
danielbeaudreau: testFile not used?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -98,17 +111,35 @@ protected void store(
...
-        destinationClient.stowRs(inWithHeader);
+        processorList.add(redactor::redact);
+      }
+
+      if(transcodeToSyntax != null && !transcodeToSyntax.equals(transferSyntax)) {
+        processorList.add((inputStream, outputStream) -> {
+          try (Transcoder transcoder = new Transcoder(inputStream)) {
+            transcoder.setIncludeFileMetaInformation(true);
+            transcoder.setDestinationTransferSyntax(transcodeToSyntax);
+            transcoder.transcode((transcoder1, dataset) -> outputStream);
```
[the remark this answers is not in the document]
red1408: It's (unused) lambda expression parameter, this is the definition itself.

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageReaderFactory.properties
```diff
@@ -0,0 +1,10 @@
+1.2.840.10008.1.2.4.50:jpeg:com.sun.imageio.plugins.jpeg.JPEGImageReader::
+1.2.840.10008.1.2.4.51:jpeg:com.sun.imageio.plugins.jpeg.JPEGImageReader::
+1.2.840.10008.1.2.4.53:jpeg:com.sun.imageio.plugins.jpeg.JPEGImageReader::
```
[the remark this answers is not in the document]
red1408: Right, .50 to .70 are all JPEG with different encoding settings, .90 and .91 are both jpeg 2000. Ideally they should be readable. But I haven't actually tested with anything other than .50 and .90 .

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageWriterFactory.properties
```diff
@@ -0,0 +1,8 @@
+1.2.840.10008.1.2.4.50:jpeg:com.sun.imageio.plugins.jpeg.JPEGImageWriter::
+1.2.840.10008.1.2.4.51:jpeg:com.sun.imageio.plugins.jpeg.JPEGImageWriter::
```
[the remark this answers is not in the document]
red1408: Yeah, I think I missed this once I got it basically functional. 
Currently this only produces 2 variants. Default ImageWriterFactory.properties in dcm4che-imageio passes extra parameters for encoding to NativeJPEGImageWriter/NativeJ2kImageWriter. Values passed are likely unique to these classes.

I could either limit to 2 writeable syntaxes for now or look into it deeper.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
red1408: This changed with dcm4che version update. Seems to do url encoding slightly differently.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
red1408: Dcm4che version update, signatures of few methods changed.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
red1408: URLEncoder.encode(value, "UTF-8") is older syntax that needs to be wrapped into try-catch,  URLEncoder.encode(value, StandardCharsets.UTF_8) doesn't need try-catch and so is cleaner.

Not really necessary to current feature, just minor refactoring.

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageWriterFactory.properties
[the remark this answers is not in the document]
danielbeaudreau: Limiting this to just .90 and .91 seems good to me. Unlikely that anyone will want to use a lossy syntax

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageReaderFactory.properties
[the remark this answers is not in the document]
danielbeaudreau: I'd say let's remove everything except .50, .90, and .91 since I'm skeptical that the JPEGImageReader can read the other obscure jpeg formats.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
danielbeaudreau: Is there any behavior change?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
danielbeaudreau: Got it thanks.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
danielbeaudreau: If not, this change is good with me

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageReaderFactory.properties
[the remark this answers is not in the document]
danielbeaudreau: You can keep RLE as well

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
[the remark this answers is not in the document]
danielbeaudreau: Got it.

danielbeaudreau:
Follow up: please move the jai image IO fork into a directory called /third_party. So the new path would be /third_party/jai-imageio-jpeg2000

integration_test/cloudbuild-integration-test.yaml
red1408: Ok

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageWriterFactory.properties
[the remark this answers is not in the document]
red1408: Ok

import/src/main/resources/com/google/cloud/healthcare/imaging/dicomadapter/transcoder/ImageReaderFactory.properties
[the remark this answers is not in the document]
red1408: Ok.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
red1408: Do you mean uploading zipped .dcm files? But they still could contain various transfer syntaxes, this seems to make more sense as a separate setting.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
red1408: Right... I've been applying transcoding to default .dcm file, which contains no single frame. Which technically passed.

Fixed this.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java
danielbeaudreau: Good point, let's leave as-is

danielbeaudreau:
Thanks, only issue remaining is combining the build steps, so we don't need to mvn install the jai code whenever we build, is that possible?

red1408:
I think I'll have to add gradle build file to jai-imageio-jpeg2000 to make that work. Getting a functional build shouldn't be a problem, but producing exactly same jar as maven probably would (with meta info and licences) if that's important.

cloudbuild.yaml
```diff
@@ -1,8 +1,24 @@
 steps:
+
+- name: ${_MAVEN_IMAGE}
```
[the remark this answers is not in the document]
red1408: Ok, added gradle build file to jai-imageio-jpeg2000.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtilTest.java
red1408: After checking it with few strings: difference is that now trailing(!) spaces are dropped when setting dcm4che string attribute value. Should be safe.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
```diff
@@ -56,14 +51,15 @@ public DicomWebClient(
   /**
    * Makes a WADO-RS call and returns the multipart response.
```
danielbeaudreau: comment is out of date: not a multipart response any more

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -124,54 +155,87 @@ private void validateParam(String value, String name) throws DicomServiceExcepti
...
+    return defaultDicomWebClient;
+  }
 
-    PipedOutputStream pdvPipeOut = new PipedOutputStream();
-    PipedInputStream pdvPipeIn = new PipedInputStream(pdvPipeOut);
-    PipedOutputStream redactedPipeOut = new PipedOutputStream();
-    PipedInputStream redactedPipeIn = new PipedInputStream(redactedPipeOut);
+  private void processStream(Executor underlyingExecutor, InputStream inputStream,
+      List<StreamProcessor> processorList) throws Throwable {
+    if (processorList.size() == 1) {
```
danielbeaudreau: Do we have any test coverage for this block? Maybe we should add a section to the integration test where we initialize the adapter without redaction / transcoding to verify the basic case still works?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
```diff
@@ -124,54 +155,87 @@ private void validateParam(String value, String name) throws DicomServiceExcepti
...
-    }
 
-    try {
-      for (int i = 0; i < 2; ++i) {
-        ecs.take().get();
+      try (pdvPipeOut) {
+        // PDVInputStream is thread-locked
+        StreamUtils.copy(inputStream, pdvPipeOut);
+      } catch (IOException e) {
+        // causes or is caused by exception in redactor.redact, no need to throw this up
```
danielbeaudreau: comment seems out of date

danielbeaudreau:
Thanks! I think building the library through gradle is fine

Can you please add a file that describes all changes made to the JAI library for future reference (e.g. commenting out  System.runFinalizersOnExit(true), changes made to make it compatible with dcm4che)

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Changed exact wording, but idea is the same.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
red1408: Ok, changed.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java
red1408: Almost all unit tests use this path (except transcoding test). There is no complex initialization that would make it more likely to fail in integration vs unit test either.
Integration test is already quite overcomplicated too...

Issue #60, opened 2020-02-25
Create GitHub Release Entry for Automated Releases


dzelemba:
This will make releases more discoverable.

danielbeaudreau:
I'm getting a not found response from Github after merging the recent PR. Have you seen this one before?

Step #29 - "publish-image": #=#=#                                                                         

######################################################################## 100.0%
######################################################################## 100.0%
Step #29 - "publish-image": {
Step #29 - "publish-image":   "message": "Not Found",
Step #29 - "publish-image":   "documentation_url": "https://developer.github.com/v3/repos/releases/#create-a-release"
Step #29 - "publish-image": }
Finished Step #29 - "publish-image"
ERROR
ERROR: build step 29 "google/cloud-sdk:272.0.0" failed: step exited with non-zero status: 1

pavertomato:
> I'm getting a not found response from Github after merging the recent PR. Have you seen this one before?
> 
> Step #29 - "publish-image": #=#=#
> 
> ######################################################################## 100.0%
> ######################################################################## 100.0%
> Step #29 - "publish-image": {
> Step #29 - "publish-image": "message": "Not Found",
> Step #29 - "publish-image": "documentation_url": "https://developer.github.com/v3/repos/releases/#create-a-release"
> Step #29 - "publish-image": }
> Finished Step #29 - "publish-image"
> ERROR
> ERROR: build step 29 "google/cloud-sdk:272.0.0" failed: step exited with non-zero status: 1

yep, looks like github api is not enabled for repo or robot user don't have permissions
https://github.com/GoogleCloudPlatform/wsi-to-dicom-converter/issues/6#issuecomment-510118455
or i may add debug line to print url to make sure what $ REPO_NAME is not changed, but since recent releases of other projects are fine, it seems not

danielbeaudreau:
Fixed now.

PR #65, opened 2020-04-01
[#60] configurable secrets

pavertomato:
since cloud build does not allow to:
- make decryption optional
- make secret decryption configurable by substitutions
- trigger one yaml configuration from another
and if user don't have access to key whole build will fail
i have workaround to make it optional: we replace yaml configuration with bash script, it doing same actions: decryption and creation of env variable, but now user may change configuration by substitutions, or if does not, build will fail only on last step

### DIFF
--- a/integration_test/cloudbuild-integration-test.yaml
+++ b/integration_test/cloudbuild-integration-test.yaml
@@ -362,9 +362,22 @@
 
 - name: ${_CLOUD_SDK_IMAGE}
   id: publish-image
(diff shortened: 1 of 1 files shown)


danielbeaudreau:
No container was created after merging this PR, could there be a bug?

pavertomato:
> No container was created after merging this PR, could there be a bug?

seems so, but in some reason without any error message, it works in my fork, may be something is wrong with access token... i'll add debug messages, may we could test it in different branch, not in master?

danielbeaudreau:
Could it be that bash -c is not capturing the return code? Debugging in a separate branch sounds good to me. Can you test that if the publish script returns an error in your fork, that the build fails?

PR #71, opened 2020-04-23
[#70] Upgrade API requests from v1beta to v1

red1408:
Since api version isn't actually hard-coded anywhere, it's just matter of updating integration test and samples.

In latest gcloud sdk version (290.0.0), healthcare is still beta, so usage via gcloud didn't change.

### DIFF
--- a/samples/data-protection-toolkit-minimal.yaml
+++ b/samples/data-protection-toolkit-minimal.yaml
@@ -71,11 +71,10 @@ projects:
                     - containerPort: 2575
                       protocol: TCP
                       name: "port"
-                  command:
-                    - "/import/bin/import"
+                  args:
                     - "--dimse_aet=IMPORTADAPTER"
                     - "--dimse_port=2575"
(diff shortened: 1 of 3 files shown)


samples/data-protection-toolkit-minimal.yaml
```diff
@@ -71,11 +71,10 @@ projects:
                     - containerPort: 2575
                       protocol: TCP
                       name: "port"
-                  command:
-                    - "/import/bin/import"
+                  args:
                     - "--dimse_aet=IMPORTADAPTER"
                     - "--dimse_port=2575"
-                    - "--dicomweb_address=https://healthcare.googleapis.com/v1beta1/projects/my-project/locations/us-central1-c/datasets/dataset/dicomStores/dicomStores/dicomWeb"
+                    - "--dicomweb_address=https://healthcare.googleapis.com/v1/projects/my-project/locations/us-central1-c/datasets/dataset/dicomStores/dicomStores/dicomWeb"
```
danielbeaudreau, line 77: We need to update the container image to 0.2.1 as well if we are switching to args.

red1408: Ok

PR #83, opened 2020-06-19
Add configuration files for Marketplace solution deployment

George-Wu:
Add configuration files that are used for building a deployment image which manages the deployment process of a Dicom Adapter.

The deployment image is used by Google Cloud Marketplace to deploy Dicom Adapter applications. This allows customers to one click deploy their own adapters on Google Cloud as well as fill in the arguments easily for their adapters through an interactive UI.

### DIFF
--- a/marketplace_confg/chart/dicom-adapter/Chart.yaml
+++ b/marketplace_confg/chart/dicom-adapter/Chart.yaml
@@ -1,2 +1,3 @@
 name: dicom-adapter-mp
 version: "0.2.1"
+apiVersion: v1
--- a/marketplace_confg/chart/dicom-adapter/Chart.yaml
+++ b/marketplace_confg/chart/dicom-adapter/Chart.yaml
@@ -1,2 +1,2 @@
 name: dicom-adapter-mp
-version: 0.2
+version: "0.2.1"
--- a/marketplace_confg/chart/dicom-adapter/Chart.yaml
+++ b/marketplace_confg/chart/dicom-adapter/Chart.yaml
@@ -1,0 +1,2 @@
+name: dicom-adapter-mp
+version: 0.2
--- a/marketplace_confg/chart/dicom-adapter/templates/application.yaml
+++ b/marketplace_confg/chart/dicom-adapter/templates/application.yaml
@@ -33,7 +33,7 @@ spec:
       DIMSE protocols (e.g., C-STORE) and the RESTful DICOMweb protocols (e.g., STOW-RS).
     maintainers:
       - name: Google HCLS Dicom team
-        email: hcls-dicom-api-team@google.com
+        email: cloud-healthcare-help@google.com
     owners:
       - name: Google HCLS Dicom team
         email: hcls-dicom-api-team@google.com
--- a/marketplace_confg/chart/dicom-adapter/templates/application.yaml
+++ b/marketplace_confg/chart/dicom-adapter/templates/application.yaml
@@ -1,0 +1,41 @@
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: "{{ .Release.Name }}"
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app.kubernetes.io/name: "{{ .Release.Name }}"
+  annotations:
+    marketplace.cloud.google.com/deploy-info: '{"partner_id": "fill in partner_id", "product_id": "fill in product_id", "partner_name": "fill in partner_name"}'
+spec:
+  selector:
+    matchLabels:
+     app.kubernetes.io/name: "{{ .Release.Name }}"
+  componentKinds:
+    - group: v1
+      kind: Service
+    - group: apps/v1
+      kind: StatefulSet
+  addOwnerRef: true
+  descriptor:
+    type: "dicom adapter"
+    version: "0.2"
+    keywords:
+      - "dicom"
+      - "adapter"
+    links:
+      - description: Github page
+        url: "https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter"
+      - description: Google health API dicom docs
+        url: "https://cloud.google.com/healthcare/docs/how-tos/dicom"
+    description: |-
+      The DICOM adapter is a set of components that translate between traditional DICOM 
+      DIMSE protocols (e.g., C-STORE) and the RESTful DICOMweb protocols (e.g., STOW-RS).
+    # need confirmation
+    maintainers:
+      - name: Google HCLS Dicom team
+        email: hcls-dicom-api-team@google.com
+    # need confirmation
+    owners:
+      - name: Google HCLS Dicom team
+        email: hcls-dicom-api-team@google.com
--- a/marketplace_confg/chart/dicom-adapter/templates/application.yaml
+++ b/marketplace_confg/chart/dicom-adapter/templates/application.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     app.kubernetes.io/name: "{{ .Release.Name }}"
   annotations:
-    marketplace.cloud.google.com/deploy-info: '{"partner_id": "fill in partner_id", "product_id": "fill in product_id", "partner_name": "fill in partner_name"}'
+    marketplace.cloud.google.com/deploy-info: '{"partner_id": "", "product_id": "", "partner_name": ""}'
 spec:
   selector:
     matchLabels:
@@ -19,7 +19,7 @@ spec:
   addOwnerRef: true
   descriptor:
     type: "dicom adapter"
-    version: "0.2"
+    version: "0.2.1"
     keywords:
       - "dicom"
       - "adapter"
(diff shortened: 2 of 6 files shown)


marketplace_confg/chart/dicom-adapter/templates/application.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - description: Github page
+        url: "https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter"
+      - description: Google health API dicom docs
+        url: "https://cloud.google.com/healthcare/docs/how-tos/dicom"
+    description: |-
+      The DICOM adapter is a set of components that translate between traditional DICOM 
+      DIMSE protocols (e.g., C-STORE) and the RESTful DICOMweb protocols (e.g., STOW-RS).
+    maintainers:
+      - name: Google HCLS Dicom team
+        email: hcls-dicom-api-team@google.com
```
danielbeaudreau, line 36: Let's use cloud-healthcare-help@google.com as the email here and below for customer support

marketplace_confg/chart/dicom-adapter/templates/application.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  selector:
+    matchLabels:
+     app.kubernetes.io/name: "{{ .Release.Name }}"
+  componentKinds:
+    - group: v1
+      kind: Service
+    - group: apps/v1
+      kind: StatefulSet
+  addOwnerRef: true
```
danielbeaudreau, line 19: What does this do?

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
```diff
@@ -0,0 +1,128 @@
...
+            {{- end }}
+        {{- end }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Release.Name }}-dicom-adapter-lb
+  {{- if .Values.lb.internal }}
+  annotations:
+    cloud.google.com/load-balancer-type: "Internal"
```
danielbeaudreau, line 114: Would this deployment work without the Internal annotation? Do we know of a use case that doesn't use Internal? if not might be fine to leave this as always internal

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
```diff
@@ -0,0 +1,128 @@
...
+kind: StatefulSet
+metadata:
+  name: "{{ .Release.Name }}-adapter"
+  namespace: "{{ .Release.Namespace }}"
+  labels: &DicomDeploymentLabels
+    app.kubernetes.io/name:  "{{ .Release.Name }}"
+    app.kubernetes.io/component: dicom-adapter-app
+spec:
+  replicas: 1
+  serviceName: {{ .Release.Name }}-dicom-adapter-lb
```
danielbeaudreau, line 11: Seems like this is the adapter config not the lb config

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
```diff
@@ -0,0 +1,128 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: "{{ .Release.Name }}-adapter"
```
danielbeaudreau, line 4: Just wondering what would this name property resolve to?

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
```diff
@@ -0,0 +1,128 @@
...
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: ENV_POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+            - name: ENV_CONTAINER_NAME
+              value: dicom-import-adapter
+            {{- end }}
+        {{- if .Values.export.hasexporter }}
```
danielbeaudreau, line 61: Were you able to test this to verify this config works?

danielbeaudreau:
Thanks Jiahao. I'm not very familiar with cloud marketplace configs, so please feel free to correct me if I made some mistakes on the code review.

marketplace_confg/chart/dicom-adapter/templates/application.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: app.k8s.io/v1beta1
+kind: Application
+metadata:
+  name: "{{ .Release.Name }}"
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app.kubernetes.io/name: "{{ .Release.Name }}"
+  annotations:
+    marketplace.cloud.google.com/deploy-info: '{"partner_id": "", "product_id": "", "partner_name": ""}'
```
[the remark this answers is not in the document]
George-Wu: We only need to fill this out when creating a new deploy image.

marketplace_confg/chart/dicom-adapter/Chart.yaml
```diff
@@ -0,0 +1,2 @@
+name: dicom-adapter-mp
```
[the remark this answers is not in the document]
George-Wu: Fixed. We are using v1 here. 
I forgot to add it because Marketplace only supports apiVersion v1 which is the default version.

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
George-Wu: Yes it will work without the internal annotation, without the internal flag users can access the adapter using a external IP.
I am not sure if there exists such a use case and I mainly expose the lb for test purposes. For example, I can use dcmsend on my computer to send dcm files to the import adapter.

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
George-Wu: The serviceName is used for pointing out the Service the adapter is going to use. In this case, it's pointing to the load balancer we define in the same file.

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
George-Wu: The .Release.Name is an instance name decided by the users when they are deploying adapter on the Marketplace.

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
George-Wu: Yes, I tested the stack driver and it logged correct events.

marketplace_confg/chart/dicom-adapter/templates/application.yaml
George-Wu: Fixed. cloud-healthcare-help@google.com is used as the maintainer's email now.

marketplace_confg/chart/dicom-adapter/templates/application.yaml
George-Wu: componentKinds is used to indicate the type of resources that the instance is using. This link provides details of the application object (required by Marketplace): https://github.com/kubernetes-sigs/application/blob/master/docs/api.md

marketplace_confg/chart/dicom-adapter/templates/application.yaml
```diff
@@ -0,0 +1,39 @@
...
+        url: "https://cloud.google.com/healthcare/docs/how-tos/dicom"
+    description: |-
+      The DICOM adapter is a set of components that translate between traditional DICOM 
+      DIMSE protocols (e.g., C-STORE) and the RESTful DICOMweb protocols (e.g., STOW-RS).
+    maintainers:
+      - name: Google HCLS Dicom team
+        email: cloud-healthcare-help@google.com
+    owners:
+      - name: Google HCLS Dicom team
+        email: hcls-dicom-api-team@google.com
```
danielbeaudreau, line 39: Let's change this one too. We want all support tickets to come through the help email address

marketplace_confg/chart/dicom-adapter/templates/application.yaml
danielbeaudreau: Thanks for the link

marketplace_confg/chart/dicom-adapter/templates/manifest.yaml
danielbeaudreau: Can we make lb.internal default to true, to match our readme recommendation and keep it more locked down by default?

George-Wu: That makes sense. Fixed.

marketplace_confg/chart/dicom-adapter/templates/application.yaml
George-Wu: fixed. cloud-healthcare-help@google.com is used as owner's email now.

Issue #77, opened 2020-06-18
Export adapter is broken by address validation


danielbeaudreau:
The export adapter validates that dicomweb_addr is a valid dicomweb address, however it isn't supposed to be. So the export adapter is broken at the current container version

Issue #84, opened 2020-06-19
Refactor integration tests


dzelemba:
Filling this as a future cleanup.

Integration tests are currently using a combination of bash scripts & cloud build settings to run a host of tests all part of one global config. This causes a few issues:
  - it's hard to isolate individual test cases to understand what is being covered
  - changing one test case can break other independent tests (e.g. having to update the build step names, or inserting into the one shared dicomstore)
  - cleanup is easy to miss
  - diff checks are just based on file comparisons which makes them hard to debug

Instead we can move towards using a proper language to write the tests (like Java or Python) that provides connivence methods for creating/cleaning up test environments. A design doc outlining this would be a good next step.

Issue #78, opened 2020-06-18
Integration test for export adapter


danielbeaudreau:
This would avoid issues like: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/issues

PR #90, opened 2020-06-28
[#77] path validation has been removed for export adapter

pavertomato:
do we actually need path validation at all? i guess import and export adapters may be used with custom endpoints and with reverse proxies/caches 

i this PR: 

- validation has been detached from dicom clients and called separately in import adapter but not in export adapter
- tests of validation has been moved to dicom_utils


dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebValidation.java
```diff
@@ -18,7 +18,8 @@
     public static ValidationPattern DICOMWEB_STUDIES_VALIDATION =
```
danielbeaudreau, line 18: Let's delete the DICOMWEB_STUDIES_VALIDATION,  STUDIES, and STUDIES_PATH variables since they aren't used any more, as well as the associated code

pavertomato:
>  Unfortunately we don't have integration tests for it atm so it's easy to break it, and this flow is broken before your PR which should fix it.

i've tested it on local machine, with custom pubsub, regarding integration tests we need to prioritize tasks https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/issues/84 and https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/issues/78

do we need to rewrite all tests and add export adapter tests, or we need to create export adapter tests and rewrite them after it?

danielbeaudreau:
Hi Egor, a simple extension to the current integration tests to add steps for an export adapter is all that is needed for now.
I think #84 is out of scope for this task, because rewriting all of the tests is a much larger effort that could take a while.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebValidation.java
pavertomato: done

pavertomato:
> Hi Egor, a simple extension to the current integration tests to add steps for an export adapter is all that is needed for now.
> I think #84 is out of scope for this task, because rewriting all of the tests is a much larger effort that could take a while.

dicom adapter integration tests is quite time consuming, i'll need a bit more time to add simple export adapter test

danielbeaudreau:
Yep that's fine. An integration test isn't needed for merging this PR

PR #91, opened 2020-07-02
Added fuzzymatching to AttributesUtil.java

satheesh09:
The current version didn't specify configs to enable fuzzymatching attribute that will perform wildcard search on CFIND Requests specifically when partial patient name (first or last name) is supplied as an input parameter. The current version works in such a way that we need to provide the exact patient full name as such it appears in the DICOM tag which would be challenging for any user to provide that as an input to a CFIND query. As per the Google documentation on fuzzymatching (https://cloud.google.com/healthcare/docs/dicom#search_parameters), I added fuzzymatching attribute to negotiate fuzzy semantic patient name attribute matching.


util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -42,6 +42,8 @@
 
   private static final int INSTANCES_LIMIT = 50000;
   private static final int STUDIES_SERIES_LIMIT = 5000;
+  
+  private static final String FUZZY_MATCHING = "true";
```
danielbeaudreau, line 46: Can we make this a boolean flag registered in https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/master/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java so that users could enable or disable it if their backend doesn't support the fuzzy matching parameter?

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
```diff
@@ -118,9 +120,19 @@ public static String attributesToQidoPath(Attributes attrs, String... includeFie
       switch (attrs.getString(Tag.QueryRetrieveLevel)) {
         case "STUDY":
           qidoPath.append("studies?limit=" + STUDIES_SERIES_LIMIT + "&");
+          
+          if (FUZZY_MATCHING == "true") {
+            qidoPath.append("fuzzymatching=true" + "&");
+          }
+
```
danielbeaudreau, line 127: We can also append fuzzy matching for an instances level retrieval, so I think this if statement could be moved outside of the switch statement, since it would apply for all query retrieve levels

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
satheesh09: Yes that's a good idea, I'll make that change and will request your review.

util/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/AttributesUtil.java
satheesh09: Actually I have tried adding it out of the switch, but I got a build failure for doing fuzzy matching at INSTANCE level. That's why I put it inside STUDY and SERIES level.

danielbeaudreau: What was the build failure? It might be a unit test that needs to be updated?

satheesh09: Yes, it is a unit test.

expected: instances?limit=50000&00080061=MG&
but was : instances?limit=50000&fuzzymatching=true&00080061=MG&
	at com.google.cloud.healthcare.imaging.dicomadapter.AttributesUtilTest.testAttributesToQidoPathArray_manyModalitiesSet(AttributesUtilTest.java:105)

danielbeaudreau: Ok, can you please update the tests as well so that the build passes with fuzzy matching enabled for all three retrieval levels?

satheesh09: Yes, I can do that!

satheesh09: As per our discussion I have added the fuzzymatching attribute to Flags and referenced it in code. Let me know if everything looks OK after you reviewed my changes.

pavertomato: could you please add unit tests to check what url is have orhave not fuzzymatching parameter on different retrieval levels?
also if you don't mind - squash commits into single one

danielbeaudreau: +1, please add a unit test

danielbeaudreau: Looks good overall though. I'm not sure why the contributor license agreement presubmit check isn't running, I'll need to figure that out before we submit this.

satheesh09: added until test for fuzzy matching.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
```diff
@@ -117,6 +120,28 @@ public JSONArray qidoRs(String path) throws DicomWebException {
...
+  
+  @Test
+  public void testCFindService_withFuzzyMatching() throws Exception {
+	cFINDFlags.fuzzyMatching = true;
+    basicCFindServiceTest(new TestUtils.DicomWebClientTestBase() {
+      @Override
+      public JSONArray qidoRs(String path) throws DicomWebException {
+        throw new DicomWebException("test-generated exception", Status.InvalidAttributeValue);
+      }
+    }, Status.InvalidAttributeValue);
```
danielbeaudreau, line 132: Since this isn't testing an error case, I think we should return a dummy response and assert for a success status in both of these tests. E.g:

 @Override
      public JSONArray qidoRs(String path) throws DicomWebException {
        JSONArray instances = new JSONArray();
        instances.put(TestUtils.dummyQidorsInstance());
        return instances;
      }
    }, Status.Success);

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
```diff
@@ -117,6 +120,28 @@ public JSONArray qidoRs(String path) throws DicomWebException {
       }
     }, Status.NotAuthorized);
   }
+  
+  @Test
+  public void testCFindService_withFuzzyMatching() throws Exception {
+	cFINDFlags.fuzzyMatching = true;
+    basicCFindServiceTest(new TestUtils.DicomWebClientTestBase() {
+      @Override
+      public JSONArray qidoRs(String path) throws DicomWebException {
```
danielbeaudreau, line 129: Can we assert that the path contains fuzzymatching=true? By using something like assertThat(path).isEqualTo("expected value") within the qidoRS function? Similarly for the other test can we assert that the path does not contain fuzzymatching=true?

satheesh09: done.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CFindServiceTest.java
satheesh09: done. I also double checked the contributor license agreement. I can find my Individual Agreements on file that I signed on Jul 02, 2020.

danielbeaudreau:
Thanks, everything looks good, but there's still some merge conflicts, could you please fix those?

satheesh09:
Conflicts resolved.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -112,8 +112,8 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
 
     // Handle C-FIND
     IDicomWebClient dicomWebClient =
-        new DicomWebClient(requestFactory, dicomwebAddress, STUDIES);
-    CFindService cFindService = new CFindService(dicomWebClient);
+        new DicomWebClient(requestFactory, flags.dicomwebAddress, STUDIES);
```
danielbeaudreau, line 115: This should be just dicomwebAddress

satheesh09: oh Okay

PR #99, opened 2020-08-05
To upstream retries pr 2 (GcpBackupUploader + HttpErrorHandling + Integration tests)

oleksii-opanasenko:
GcpBackupUploader with tests added;
Http Errors filter added + Handling with tests added;
Integration tests added;


danielbeaudreau:
I just merged the other PR to the retries branch, could you please resolve the merge conflicts and then I'll review this one? Thanks

oleksii-opanasenko:
> I just merged the other PR to the retries branch, could you please resolve the merge conflicts and then I'll review this one? Thanks

Done

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
```diff
@@ -1,25 +1,129 @@
...
+  }
+
+  public String getUploadObject() {
+    return uploadObject;
+  }
+
+  public Credentials getCredential(String env) throws IOException {
+    return GoogleCredentials
+            .fromStream(new FileInputStream(System.getenv(env)));
+  }
```
danielbeaudreau, line 108: We get credentials in the export adapter here: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/e72628beaf645b70264f377d782a8b6315c8f00c/export/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ExportAdapter.java#L56
Could we use the same method here? to avoid relying on an env variable

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
```diff
@@ -1,25 +1,129 @@
...
+import com.google.cloud.storage.BlobId;
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.channels.Channels;
+import java.util.List;
+import org.apache.http.client.utils.URIBuilder;
```
danielbeaudreau, line 16: I'd prefer to minimize the use of external libraries here, because the more we import the more likely that we get a flaky build. I think it should be easy to mimic the functionality of URIBuilder here?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/monitoring/Event.java
```diff
@@ -21,6 +21,8 @@
   CSTORE_ERROR(Constants.prefix + "cstore_errors"),
   CSTORE_BACKUP_ERROR(Constants.prefix + "cstore_backup_errors"),
   CSTORE_BYTES(Constants.prefix + "cstore_bytes"),
+  CSTORE_409_ERROR(Constants.prefix + "cstore_409_error"),
+  CSTORE_5xx_ERROR(Constants.prefix + "cstore_5xx_error"),
```
danielbeaudreau, line 25: We already have monitoring codes for CSTORE_ERROR, adding a metric for every status code might make this hard to configure alerts on. Can we just increment the regular CSTORE_ERROR metric for all additional codes?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
```diff
@@ -1,25 +1,129 @@
...
+      validatePathParameter(uploadFilePath, "upload file path");
+      List<String> segments = new URIBuilder().setPath(getUploadFilePath()).getPathSegments();
+      projectName = segments.get(2);
+      bucketName = segments.get(3);
+      uploadObject = new URIBuilder()
+              .setPathSegments(segments.subList(4, segments.size()))
+              .getPath().substring(1);
+      validatePathParameter(projectName, "project name");
+      validatePathParameter(bucketName, "bucket name");
+      validatePathParameter(uploadObject, "upload object");
```
danielbeaudreau, line 87: Isn't it more correct to call this the folder rather than the upload object?

Also, traditionally GCS file paths are "gs://bucketID/path/to/object", the project ID isn't included in the GCS URI. Could we split the GCS backup project ID into a separate flag (something like --gcs_backup_project_id ) so we can re-use the established concept of the GCS URI without any changes here?  The readme would also need to be updated.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
```diff
@@ -1,25 +1,129 @@
...
+      storage.delete(bucketName, getFullUploadObject(uniqueFileName));
+    } catch (Exception e) {
+      throw new BackupException("Error with removing backup file: " + e.getMessage(), e);
+    }
+  }
+
+  private void parseUploadFilePath(String uploadFilePath) throws GcpUriParseException {
+    try {
+      validatePathParameter(uploadFilePath, "upload file path");
+      List<String> segments = new URIBuilder().setPath(getUploadFilePath()).getPathSegments();
```
danielbeaudreau, line 79: Let's also validate that this starts with "gs://"

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
NikolaiKovalenko: The documentation notes that the GCS structure includes a bucket and objects, the concept of a folder is not suitable for GCS -> https://cloud.google.com/storage/docs/introduction

danielbeaudreau: Ok what about 'object path prefix' or something like that?

NikolaiKovalenko: it looks like this is the name, not the path to it, because the name of the object may look like /level1/level2/some_file -> https://cloud.google.com/storage/docs/json_api/v1/objects

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -407,10 +514,18 @@
...
   - delete-dicom-store
 
+- name: ${_CLOUD_SDK_IMAGE}
+  id: delete-backup-bucket
+  args:
+  - gsutil
+  - rm
+  - "-r"
+  - "gs://${_BUCKET}"
+
```
danielbeaudreau, line 524: I think we're going to want to use a fixed bucket here instead, because we don't want to grant the cloud build service account broad scope gcs permissions on the test project. So could you change the tests to assume that the bucket already exists, and I will make sure that we have the right permissions to access it?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploaderTest.java
```diff
@@ -0,0 +1,209 @@
...
+    @BeforeClass
+    public static void setUp() {
+        try {
+            RemoteStorageHelper helper =
+                    RemoteStorageHelper.create(PROJECT_NAME, new FileInputStream(System.getenv(ENV_CREDS)));
+            localStorage = helper.getOptions().getService();
+            localStorage.create(BucketInfo.of(BUCKET_NAME));
+            gcpBackupUploader = new GcpBackupUploader(UPLOAD_PATH, localStorage);
+        } catch (IOException e) {
+            e.printStackTrace();
```
danielbeaudreau, line 61: Since this project is open source, anyone should be able to run the unit tests. I think dialing into a remote storage on a cloud project won't really work for us, e.g. imagine two users run these tests in parallel, they may fail. Also this requires us to give all users access to a remote project and bucket. Can we re-work these tests so that they don't require using remote GCS storage? 

One solution could be to make the backup uploader take in an interface for interacting with GCS, and then mock out that interface ourselves.

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
```diff
@@ -334,16 +345,58 @@ public void testBackupUploadService_ExceptionOnWriteToBackup_bytesToStowRsNotAva
...
+
+    basicCStoreServiceTest(
+        Status.Success,
+        backupUploadService,
+        spyStowClient);
+
+    verify(mockBackupUploader, timeout(100).times(1)).doWriteBackup(any(InputStream.class), anyString());
+    verify(mockBackupUploader, timeout(100).times(3)).doReadBackup(anyString());
+    verify(spyStowClient, timeout(100).times(3)).stowRs(any(InputStream.class));
+    verify(mockBackupUploader, timeout(100).times(1)).doRemoveBackup(anyString());
```
danielbeaudreau, line 392: Let's avoid using timeout to reduce the chance of flaky test runs

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploaderTest.java
NikolaiKovalenko: This approach to create Storage was used on the recommendation of Google developers -> https://github.com/googleapis/java-storage#quickstart. It is impossible to use the LocalStorageHelper class for our tests, since it does not implement the function of emulating data writing. Immediately after the end of the tests, the temporary Bucket is deleted and will be created again for the next user. This gives us a guarantee of the correct operation and the deletion of temporary data

NikolaiKovalenko: As a solution, I propose to use the option of creating a Storage object using the default settings, the user will not need to create ENVs and do not need to specify credentials, also Bucket id will be generated randomly based on UUIDs, where the user can run tests at the same time

danielbeaudreau: Hmm, but I also don't think we want this project to depend on a bucket in an external project, and give users permissions to write to that project. They could write whatever they want to those buckets with the credentials right?

NikolaiKovalenko: All tests are executed in a temporary Bucket, which is removed immediately. No existing Bucket will be affected or changed. Also, the user does not know the name of the Bucket and does not affect the process of its creation, modification or allocation.There is one bottleneck - the user must have write permissions. But if he uses the option with gcp backup, then he must have such permissions for the backup to work.

danielbeaudreau: Yeah, but still the user can modify this code to write whatever they want to your bucket right? Also linking this project with an external bucket and leaving the credentials in the text file is a bit worrisome. And if we don't leave the credentials in the file, no one will be able to run this test.

I'm fine with deleting this test altogether rather than using an external bucket and remote storage, but I think creating an interface for interacting with GCS, similar to how we create a DicomWebClient and mock it in other tests, would be the ideal approach.

oleksii-opanasenko: local FakeStorage implemented by Nikolai

oleksii-opanasenko:
Well, i think we done all necessary changes. @danielbeaudreau  please review last changes.

oleksii-opanasenko:
Well, i think we done all necessary changes. @danielbeaudreau  please review last changes.

README.md
```diff
@@ -299,8 +299,8 @@ The following arguments are used to configure the mode:
 
 If the flag --persistent_file_storage_location is not used then loading occurs without writing a temporary file.
 
-if the value for the flag --persistent_file_storage_location is specified in the format gs://project-id/bucket-id/some-object,
- the file will be written to the specified project bucket. It is also necessary to set the system variable [GOOGLE_APPLICATION_CREDENTIALS](https://cloud.google.com/healthcare/docs/how-tos/authentication#setting_the_environment_variables) = path_to_file_with_json_data.
+if the value for the flag --persistent_file_storage_location is specified in the format gs://bucket-id/some-object,
```
danielbeaudreau, line 302: I think it would be more accurate to say "gs://bucket-id/some-directory" rather than "gs://bucket-id/some-object", since this is really a folder where we're writing the files to. Some of our official docs already use this 'directory' terminology: https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.fhirStores/import#gcssource

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadServiceTest.java
```diff
@@ -179,17 +185,71 @@ public void startUploading_noMoreTry_BackupException() throws BackupException {
   public void startUploading_lastAttempt_noNewSchedule_BackupException() throws IOException, DicomWebException {
     spyBackupState = new BackupState(UNIQUE_FILE_NAME, ATTEMPTS_AMOUNT_ONE);
 
-    doThrow(new DicomWebException("Reason2")).doNothing().when(webClientMock).stowRs(any(InputStream.class));
-    when(delayCalculatorMock.getExponentialDelayMillis(anyInt())).thenReturn(0L);
+    doThrow(new DicomWebException("Reason2", HttpStatus.INTERNAL_SERVER_ERROR_500, HttpStatusCodes.STATUS_CODE_SERVER_ERROR))
```
danielbeaudreau, line 188: The third argument to DicomWebException should be Status.ProcessingFailure since it's a DIMSE code not HTTP

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadServiceTest.java
```diff
@@ -179,17 +185,71 @@ public void startUploading_noMoreTry_BackupException() throws BackupException {
...
 
-    backupUploadService.startUploading(webClientMock, spyBackupState);
+    verify(backupUploaderMock).doReadBackup(eq(UNIQUE_FILE_NAME));
+    verify(webClientMock).stowRs(any(InputStream.class));
+    verify(backupUploaderMock, never()).doRemoveBackup(eq(UNIQUE_FILE_NAME));
+  }
+
+  @Test
+  public void startUploading_DicomWebException409Code_failed() throws DicomWebException, IOException {
+    doThrow(new DicomWebException("conflictTestCode409", HttpStatus.CONFLICT_409, HttpStatusCodes.STATUS_CODE_BAD_REQUEST))
```
danielbeaudreau, line 201: Same here

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadServiceTest.java
```diff
@@ -179,17 +185,71 @@ public void startUploading_noMoreTry_BackupException() throws BackupException {
...
+
+    verify(backupUploaderMock).doReadBackup(eq(UNIQUE_FILE_NAME));
+    verify(webClientMock).stowRs(any(InputStream.class));
     verify(backupUploaderMock, never()).doRemoveBackup(eq(UNIQUE_FILE_NAME));
   }
+
+  @Test
+  public void startUploading_DicomWebException500CodeThen501Code_noMoreTry_BackupException() throws IOException, DicomWebException {
+      doThrow(new DicomWebException("testCode500", HttpStatus.INTERNAL_SERVER_ERROR_500, HttpStatusCodes.STATUS_CODE_SERVER_ERROR))
+        .doThrow(new DicomWebException("testCode502", HttpStatus.BAD_GATEWAY_502, HttpStatusCodes.STATUS_CODE_BAD_GATEWAY))
```
danielbeaudreau, line 216: Same here

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadService.java
```diff
@@ -100,25 +110,36 @@ private void scheduleUploadWithDelay(IDicomWebClient webClient, BackupState back
...
   }
 
-  private void throwOnNoResendAttemptsLeft(String uniqueFileName) throws CompletionException {
-    throw new CompletionException(getNoResendAttemptLeftException(uniqueFileName));
+  private void throwOnNoResendAttemptsLeft(int dicomStatus, String uniqueFileName) throws CompletionException {
+    throw new CompletionException(getNoResendAttemptLeftException(dicomStatus, uniqueFileName));
+  }
+
+  private void throwOnHttpFilterFail(int dicomStatus, String uniqueFileName) throws CompletionException {
+    String errorMessage = "sopInstanceUID=" + uniqueFileName + ". Filtered by httpCode.";
```
danielbeaudreau, line 130: Can we reword this slightly to say "Not retried due to HTTP code" to make it more clear?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadService.java
```diff
@@ -100,25 +110,36 @@ private void scheduleUploadWithDelay(IDicomWebClient webClient, BackupState back
         throw new IBackupUploader.BackupException(eex.getCause());
       }
     } else {
-      throw getNoResendAttemptLeftException(uniqueFileName);
+      throw getNoResendAttemptLeftException(null, uniqueFileName);
```
danielbeaudreau, line 113: Can we pass along the original exception message in No Resend attempts left and throwOnHttpFilterFail so that no details are lost?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/FakeStorage.java
```diff
@@ -0,0 +1,455 @@
...
+    public boolean delete(String bucket, BucketSourceOption... options) {
+        return false;
+    }
+
+    @Override
+    public boolean delete(String bucket, String blob, BlobSourceOption... options) {
+        try {
+            fakeObjects.remove(Paths.get(bucket, blob).toString());
+            return true;
+        } catch (Exception igrored){
```
danielbeaudreau, line 149: typo

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploaderTest.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final String UPLOAD_PATH_EMPTY_UPLOAD_OBJECT = "gs:///some//";
+    private static final String UPLOAD_PATH_SPACE_UPLOAD_OBJECT = "gs://some/ ";
+    private static final String UPLOAD_OBJECT = "test-backup";
+
+    /**
+    * To create an instance of the Storage class, the tests used RemoteStorageHelper
+    * because LocalStorageHelper returns an "method not implemented yet" error when
+    * trying to write data to the bucket. The tests use a temporary bucket with random
+    * id that is deleted after all tests have run
+     */
```
danielbeaudreau, line 45: The comment is out of date

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploaderTest.java
```diff
@@ -0,0 +1,190 @@
...
+    @BeforeClass
+    public static void setUp() {
+        try {
+            BUCKET_NAME = RemoteStorageHelper.generateBucketName();
+            UPLOAD_PATH = "gs://".concat(BUCKET_NAME).concat("/test-backup");
```
danielbeaudreau, line 50: We could just use a constant bucket name now

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -433,8 +529,10 @@ substitutions:
 # Update to correct value if adding/removing steps before adapter run.
   _ADAPTER_RUN_STEP: 'step_5'
   _ADAPTER_RUN_STEP_HTTP2: 'step_6'
+  _ADAPTER_RUN_STEP_WITH_LOCAL_BACKUP: 'step_7'
+  _ADAPTER_RUN_STEP_WITH_GCS_BACKUP: 'step_8'
   _STORE_SCP_RUN_STEP: 'step_3'
-  _COMMITMENT_SCU_STEP: 'step_18'
+  _COMMITMENT_SCU_STEP: 'step_25'
```
danielbeaudreau, line 535: Looks like there are actually only 6 new steps rather than 7, so this should be 24 rather than 25. I might have miscounted before or this file changed.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/BackupUploadService.java
oleksii-opanasenko: We do not miss anything. In normal workflow this check is never happens, because we have outer check 
attempts > 0 in CStoreService. This check presented like an double check for future usage/reusage of BackupUploadService.
I can remove it if you want.

danielbeaudreau: I see, yes you are right for this particular exception, and keeping it as-is is fine.

But for the calls to 
throwOnNoResendAttemptsLeft(dicomStatus, uniqueFileName);
and 
throwOnHttpFilterFail(dicomStatus, uniqueFileName);
above, within the CompletableFuture.Async call,  it would be nice to include the original exception stacktrace in these messages.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -24,6 +24,7 @@
 import com.google.cloud.healthcare.deid.redactor.protos.DicomConfigProtos;
 import com.google.cloud.healthcare.deid.redactor.protos.DicomConfigProtos.DicomConfig;
 import com.google.cloud.healthcare.deid.redactor.protos.DicomConfigProtos.DicomConfig.TagFilterProfile;
+import com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.BackupFlags;
 import com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.DelayCalculator;
 import com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.GcpBackupUploader;
 import com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.IBackupUploadService;
```
danielbeaudreau, line 30: import org.apache.commons.lang3.StringUtils; can be removed

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/backupuploader/GcpBackupUploader.java
```diff
@@ -17,7 +17,7 @@
 public class GcpBackupUploader extends AbstractBackupUploader {
```
danielbeaudreau, line 17: import org.apache.commons.lang3.StringUtils; can be removed

danielbeaudreau:
We're getting a runtime error in the integration test when initializing monitoring:

Already have image: gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-import:49c453a
Exception in thread "main" java.lang.AbstractMethodError: Receiver class com.google.api.gax.grpc.InstantiatingGrpcChannelProvider does not define or inherit an implementation of the resolved method 'abstract com.google.api.gax.rpc.TransportChannelProvider withExecutor(java.util.concurrent.Executor)' of interface com.google.api.gax.rpc.TransportChannelProvider.
	at com.google.api.gax.rpc.ClientContext.create(ClientContext.java:140)
	at com.google.cloud.monitoring.v3.stub.GrpcMetricServiceStub.create(GrpcMetricServiceStub.java:176)
	at com.google.cloud.monitoring.v3.stub.MetricServiceStubSettings.createStub(MetricServiceStubSettings.java:188)
	at com.google.cloud.monitoring.v3.MetricServiceClient.<init>(MetricServiceClient.java:157)
	at com.google.cloud.monitoring.v3.MetricServiceClient.create(MetricServiceClient.java:138)
	at com.google.cloud.monitoring.v3.MetricServiceClient.create(MetricServiceClient.java:129)
	at com.google.cloud.healthcare.imaging.dicomadapter.monitoring.MonitoringService.<init>(MonitoringService.java:63)
	at com.google.cloud.healthcare.imaging.dicomadapter.monitoring.MonitoringService.initialize(MonitoringService.java:112)
	at com.google.cloud.healthcare.imaging.dicomadapter.ImportAdapter.main(ImportAdapter.java:84)



Researching this a bit, it looks like something to do with pulling in multiple GCP dependencies: 
https://github.com/firebase/firebase-admin-java/issues/447

Could you please take a look and see if you can resolve the issue, maybe by tweaking the version of the imported libraries? To reproduce locally, try specifying the monitoring_project_id when starting the adapter: 

https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/e2aec95507a4cc1bb249f211ebf7a97acfd82c50/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java#L96

danielbeaudreau:
In particular, using the 'bom' package for both monitoring and GCS dependencies might be the fix here: https://github.com/firebase/firebase-admin-java/issues/447#issuecomment-655148146

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -420,6 +515,7 @@ substitutions:
   _LOCATION: us-central1
   _DATASET: healthcare-dicom-dicomweb-adapter-test
   _STORE_NAME: integration-test-store
+  _BUCKET: healthcare-dicom-dicomweb-adapter-integration-test
```
danielbeaudreau, line 518: Please change this to a different bucket name like "dicomweb-adapter-integration-test" so that I can create the bucket in our integration test project.

oleksii-opanasenko:
> We're getting a runtime error in the integration test when initializing monitoring:
> 
> Already have image: gcr.io/cloud-healthcare-containers/healthcare-api-dicom-dicomweb-adapter-import:49c453a
> Exception in thread "main" java.lang.AbstractMethodError: Receiver class com.google.api.gax.grpc.InstantiatingGrpcChannelProvider does not define or inherit an implementation of the resolved method 'abstract com.google.api.gax.rpc.TransportChannelProvider withExecutor(java.util.concurrent.Executor)' of interface com.google.api.gax.rpc.TransportChannelProvider.
> at com.google.api.gax.rpc.ClientContext.create(ClientContext.java:140)
> at com.google.cloud.monitoring.v3.stub.GrpcMetricServiceStub.create(GrpcMetricServiceStub.java:176)
> at com.google.cloud.monitoring.v3.stub.MetricServiceStubSettings.createStub(MetricServiceStubSettings.java:188)
> at com.google.cloud.monitoring.v3.MetricServiceClient.(MetricServiceClient.java:157)
> at com.google.cloud.monitoring.v3.MetricServiceClient.create(MetricServiceClient.java:138)
> at com.google.cloud.monitoring.v3.MetricServiceClient.create(MetricServiceClient.java:129)
> at com.google.cloud.healthcare.imaging.dicomadapter.monitoring.MonitoringService.(MonitoringService.java:63)
> at com.google.cloud.healthcare.imaging.dicomadapter.monitoring.MonitoringService.initialize(MonitoringService.java:112)
> at com.google.cloud.healthcare.imaging.dicomadapter.ImportAdapter.main(ImportAdapter.java:84)
> 
> Researching this a bit, it looks like something to do with pulling in multiple GCP dependencies:
> [firebase/firebase-admin-java#447](https://github.com/firebase/firebase-admin-java/issues/447)
> 
> Could you please take a look and see if you can resolve the issue, maybe by tweaking the version of the imported libraries? To reproduce locally, try specifying the monitoring_project_id when starting the adapter:
> 
> https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/e2aec95507a4cc1bb249f211ebf7a97acfd82c50/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/Flags.java#L96

done

danielbeaudreau:
Hi Oleksii,

The integration test is failing because _ADAPTER_PORT_WITH_LOCAL_BACKUP and _ADAPTER_PORT_WITH_GCS_BACKUP variables are not defined in the yaml.

danielbeaudreau:
We also need to specify the flag gcs_backup_project_id=_PROJECT on the GCS backup adapter

NikolaiKovalenko:
flag gcs_backup_project_id was added

danielbeaudreau:
It looks like one of the new unit tests is flaky, could you please take a look?
https://pantheon.corp.google.com/cloud-build/builds/f1812b6d-f87d-487d-9e67-fe53a01533b1;step=0?project=gcp-healthcare-oss-test

"com.google.cloud.healthcare.imaging.dicomadapter.CStoreServiceTest > testBackupUploadService_ExceptionOnWriteToBackup FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at CStoreServiceTest.java:347"

danielbeaudreau:
From integration test step 8 

Exception in thread "main" com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.IBackupUploader$BackupException: Invalid upload path, parameter - oauthScopes is blank.
	at com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.AbstractBackupUploader.validatePathParameter(AbstractBackupUploader.java:17)
	at com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.GcpBackupUploader.getCredential(GcpBackupUploader.java:106)
	at com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.GcpBackupUploader.getStorage(GcpBackupUploader.java:119)
	at com.google.cloud.healthcare.imaging.dicomadapter.backupuploader.GcpBackupUploader.<init>(GcpBackupUploader.java:35)
	at com.google.cloud.healthcare.imaging.dicomadapter.ImportAdapter.configureBackupUploadService(ImportAdapter.java:155)
	at com.google.cloud.healthcare.imaging.dicomadapter.ImportAdapter.main(ImportAdapter.java:116)

oleksii-opanasenko:
@danielbeaudreau is '--oauth_scopes=https://www.googleapis.com/auth/cloud-platform' flag correct for integration test purposes ?

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
```diff
@@ -392,7 +392,7 @@ public void testBackupUploadService_retryOn_DicomWebException408Code_Success() t
 
     verify(mockBackupUploader).doWriteBackup(any(InputStream.class), anyString());
     verify(mockBackupUploader, times(2)).doReadBackup(anyString());
-    verify(spyStowClient, times(2)).stowRs(any(InputStream.class));
+    verify(spyStowClient, atLeast(1)).stowRs(any(InputStream.class));
```
danielbeaudreau, line 395: Why would this ever happen only once?

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -420,11 +517,14 @@ substitutions:
   _LOCATION: us-central1
   _DATASET: healthcare-dicom-dicomweb-adapter-test
   _STORE_NAME: integration-test-store
+  _BUCKET: dicomweb-import-adapter-integration-test
```
danielbeaudreau, line 520: I have not created this bucket yet, so the integration test shouldn't pass, but it does. Are we not propagating back errors to the store SCU user when we fail to write to a GCS bucket? 

Also once we find out why this is passing, I think we should store different example files for each adapter, and add diff steps for the extra files. Because if the input file is a duplicate the server doesn't do anything.
Meaning that we should have example2.dcm and example3.dcm in steps run-store-scu-with-local-backup and run-store-scu-with-gcs-backup and then extra check-store-curl and check-store-diff commands for each of them. This will also affect the step numbers at the bottom of the file

import/src/test/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreServiceTest.java
oleksii-opanasenko: Yes, you right, my mistake. I thought that it`s the same case that in testBackupUploadService_ExceptionOnWriteToBackup where Exception thrown and stowRs call is processed in different threads. Reverted.

integration_test/cloudbuild-integration-test.yaml
NikolaiKovalenko: We tested the GCS backup on our project and everything went well. Perhaps the error is encapsulated in the docker container and not translated out? Is it possible to check the container logs?

danielbeaudreau: Yeah I guess the store scu command doesn't return a non-zero error code on storage failure so the test won't fail. This highlights the need for storing separate files.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -387,6 +387,26 @@
...
+    - integration_test/downloaded2.dcm
+  waitFor:
+    - run-store-scu-with-local-backup
+
+- name: ${_CLOUD_SDK_IMAGE}
+  id: check-store-curl-with-gcs-backup
+  args:
+    - bash
+    - integration_test/scripts/curl-dcm.sh
+    - "https://healthcare.googleapis.com/${_VERSION}/projects/${_PROJECT}/locations/${_LOCATION}/datasets/${_DATASET}/dicomStores/${_STORE_NAME}/dicomWeb/studies/${_REPLACED_UID}/series/${_REPLACED_UID}/instances/${_REPLACED_UID}"
```
danielbeaudreau, line 405: We should use the real study, series, and instance UIDs here instead of REPLACED_UID. The UIDs will not be replaced because the two duplicate adapters do not specify any redaction flags.

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -416,6 +437,24 @@
   waitFor:
   - check-store-curl-destination-2
 
+- name: 'busybox:latest'
+  id: check-store-diff-with-local-backup
+  args:
+    - diff
+    - 'integration_test/downloaded2.dcm'
+    - 'integration_test/data/example-redacted-jp2k.dcm'
```
danielbeaudreau, line 445: Should be diffed with    integration_test/data/example2.dcm, and example3.dcm for the other step, because as mentioned we do not redact uids here.

danielbeaudreau (changes requested):

integration_test/cloudbuild-integration-test.yaml
```diff
@@ -563,7 +563,7 @@ substitutions:
   _STORE_SCP_PORT: '2576'
   _ADAPTER_PORT_HTTP2: '2577'
   _CLOSE_STORE_SCP_PORT: '3001'
-  _COMMITMENT_SCU_PORT: '4000'
+  _COMMITMENT_SCU_PORT: '4000'dicomweb-import-adapter-integration-test
```
danielbeaudreau, line 566: typo

danielbeaudreau:
From test result:
Already have image: busybox:latest
Files integration_test/downloaded3.dcm and integration_test/data/example-redacted-jp2k.dcm differ

Are you able to run these tests locally?

NikolaiKovalenko:
We've updated the file mapping in the diff steps

Perhaps launching on a local machine will not be indicative, because a lot is connected with setting up the current user and the project. We will try to do it

danielbeaudreau:
Ok, please check and fix the test until the presubmit passes, it should run automatically after each commit

danielbeaudreau:
I compared the files, found the diff between downloaded2.dcm and example2.dcm:
12c12
< (0002,0013) SH [dcm4che-3.3.8]                          #  14, 1 ImplementationVersionName
---
> (0002,0013) SH [dcm4che-5.18.1]                         #  14, 1 ImplementationVersionName

danielbeaudreau:
I think the store SCP is modifying the implementation version name, causing our problem here.
I'll send you versions of example2.dcm and example3.dcm which should fix the problem

NikolaiKovalenko:
Thank you, maybe this is true, because now downloaded2.dcm and example2.dcm files should be identical. Perhaps we should use the same _REPLACES_UID value for all example files?

danielbeaudreau:
I sent you the new files via email. Yeah, it seems the implementation version name is changing somewhere, probably in the adapter.

Not sure why it does not change for example.dcm, probably because we're using the redact library which is built in dcm4che3, rather than 5, which our adapter is using.

Using different uids is important for example2.dcm and example3.dcm because the healthcare API doesn't write the file if it's an exact duplicate. So we aren't testing anything in that case, since whether or not the file upload succeeds, the end result is the same.

NikolaiKovalenko:
new files added into repository

danielbeaudreau:
File diffs look good now. We still have to update the find scu golden files.
See step 35: 
https://console.cloud.google.com/cloud-build/builds/9e9c5fb1-92e9-4356-a4c9-d0ee7faa3b0b?project=301466572187
Please update integration_test/findscu-instance-result1.xml to contain the following XML:

<?xml version="1.0" encoding="UTF-8"?><NativeDicomModel xml:space="preserve"><DicomAttribute keyword="SOPClassUID" tag="00080016" vr="UI"><Value number="1">1.2.840.10008.5.1.4.1.1.7</Value></DicomAttribute><DicomAttribute keyword="SOPInstanceUID" tag="00080018" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014056</Value></DicomAttribute><DicomAttribute keyword="StudyDate" tag="00080020" vr="DA"/><DicomAttribute keyword="StudyTime" tag="00080030" vr="TM"/><DicomAttribute keyword="AccessionNumber" tag="00080050" vr="SH"/><DicomAttribute keyword="Modality" tag="00080060" vr="CS"><Value number="1">XC</Value></DicomAttribute><DicomAttribute keyword="ReferringPhysicianName" tag="00080090" vr="PN"/><DicomAttribute keyword="PatientName" tag="00100010" vr="PN"/><DicomAttribute keyword="PatientID" tag="00100020" vr="LO"/><DicomAttribute keyword="PatientBirthDate" tag="00100030" vr="DA"/><DicomAttribute keyword="PatientSex" tag="00100040" vr="CS"/><DicomAttribute keyword="StudyInstanceUID" tag="0020000D" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014056</Value></DicomAttribute><DicomAttribute keyword="SeriesInstanceUID" tag="0020000E" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014056</Value></DicomAttribute><DicomAttribute keyword="StudyID" tag="00200010" vr="SH"/><DicomAttribute keyword="InstanceNumber" tag="00200013" vr="IS"><Value number="1">1</Value></DicomAttribute><DicomAttribute keyword="Rows" tag="00280010" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="Columns" tag="00280011" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="BitsAllocated" tag="00280100" vr="US"><Value number="1">8</Value></DicomAttribute></NativeDicomModel><?xml version="1.0" encoding="UTF-8"?><NativeDicomModel xml:space="preserve"><DicomAttribute keyword="SOPClassUID" tag="00080016" vr="UI"><Value number="1">1.2.840.10008.5.1.4.1.1.7</Value></DicomAttribute><DicomAttribute keyword="SOPInstanceUID" tag="00080018" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014222</Value></DicomAttribute><DicomAttribute keyword="StudyDate" tag="00080020" vr="DA"/><DicomAttribute keyword="StudyTime" tag="00080030" vr="TM"/><DicomAttribute keyword="AccessionNumber" tag="00080050" vr="SH"/><DicomAttribute keyword="Modality" tag="00080060" vr="CS"><Value number="1">XC</Value></DicomAttribute><DicomAttribute keyword="ReferringPhysicianName" tag="00080090" vr="PN"/><DicomAttribute keyword="PatientName" tag="00100010" vr="PN"/><DicomAttribute keyword="PatientID" tag="00100020" vr="LO"/><DicomAttribute keyword="PatientBirthDate" tag="00100030" vr="DA"/><DicomAttribute keyword="PatientSex" tag="00100040" vr="CS"/><DicomAttribute keyword="StudyInstanceUID" tag="0020000D" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014222</Value></DicomAttribute><DicomAttribute keyword="SeriesInstanceUID" tag="0020000E" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014222</Value></DicomAttribute><DicomAttribute keyword="StudyID" tag="00200010" vr="SH"/><DicomAttribute keyword="InstanceNumber" tag="00200013" vr="IS"><Value number="1">1</Value></DicomAttribute><DicomAttribute keyword="Rows" tag="00280010" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="Columns" tag="00280011" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="BitsAllocated" tag="00280100" vr="US"><Value number="1">8</Value></DicomAttribute></NativeDicomModel><?xml version="1.0" encoding="UTF-8"?><NativeDicomModel xml:space="preserve"><DicomAttribute keyword="SOPClassUID" tag="00080016" vr="UI"><Value number="1">1.2.840.10008.5.1.4.1.1.7</Value></DicomAttribute><DicomAttribute keyword="SOPInstanceUID" tag="00080018" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014333</Value></DicomAttribute><DicomAttribute keyword="StudyDate" tag="00080020" vr="DA"/><DicomAttribute keyword="StudyTime" tag="00080030" vr="TM"/><DicomAttribute keyword="AccessionNumber" tag="00080050" vr="SH"/><DicomAttribute keyword="Modality" tag="00080060" vr="CS"><Value number="1">XC</Value></DicomAttribute><DicomAttribute keyword="ReferringPhysicianName" tag="00080090" vr="PN"/><DicomAttribute keyword="PatientName" tag="00100010" vr="PN"/><DicomAttribute keyword="PatientID" tag="00100020" vr="LO"/><DicomAttribute keyword="PatientBirthDate" tag="00100030" vr="DA"/><DicomAttribute keyword="PatientSex" tag="00100040" vr="CS"/><DicomAttribute keyword="StudyInstanceUID" tag="0020000D" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014333</Value></DicomAttribute><DicomAttribute keyword="SeriesInstanceUID" tag="0020000E" vr="UI"><Value number="1">2.25.140302709094137852884202099990798014333</Value></DicomAttribute><DicomAttribute keyword="StudyID" tag="00200010" vr="SH"/><DicomAttribute keyword="InstanceNumber" tag="00200013" vr="IS"><Value number="1">1</Value></DicomAttribute><DicomAttribute keyword="Rows" tag="00280010" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="Columns" tag="00280011" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="BitsAllocated" tag="00280100" vr="US"><Value number="1">8</Value></DicomAttribute></NativeDicomModel><?xml version="1.0" encoding="UTF-8"?><NativeDicomModel xml:space="preserve"><DicomAttribute keyword="SOPClassUID" tag="00080016" vr="UI"><Value number="1">1.2.840.10008.5.1.4.1.1.7</Value></DicomAttribute><DicomAttribute keyword="SOPInstanceUID" tag="00080018" vr="UI"><Value number="1">111</Value></DicomAttribute><DicomAttribute keyword="StudyDate" tag="00080020" vr="DA"/><DicomAttribute keyword="StudyTime" tag="00080030" vr="TM"/><DicomAttribute keyword="AccessionNumber" tag="00080050" vr="SH"/><DicomAttribute keyword="Modality" tag="00080060" vr="CS"><Value number="1">XC</Value></DicomAttribute><DicomAttribute keyword="ReferringPhysicianName" tag="00080090" vr="PN"/><DicomAttribute keyword="PatientName" tag="00100010" vr="PN"/><DicomAttribute keyword="PatientID" tag="00100020" vr="LO"/><DicomAttribute keyword="PatientBirthDate" tag="00100030" vr="DA"/><DicomAttribute keyword="PatientSex" tag="00100040" vr="CS"/><DicomAttribute keyword="StudyInstanceUID" tag="0020000D" vr="UI"><Value number="1">111</Value></DicomAttribute><DicomAttribute keyword="SeriesInstanceUID" tag="0020000E" vr="UI"><Value number="1">111</Value></DicomAttribute><DicomAttribute keyword="StudyID" tag="00200010" vr="SH"/><DicomAttribute keyword="InstanceNumber" tag="00200013" vr="IS"><Value number="1">1</Value></DicomAttribute><DicomAttribute keyword="Rows" tag="00280010" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="Columns" tag="00280011" vr="US"><Value number="1">500</Value></DicomAttribute><DicomAttribute keyword="BitsAllocated" tag="00280100" vr="US"><Value number="1">8</Value></DicomAttribute></NativeDicomModel>


then please push a commit and do the same for the other find scu diff steps to update their expected XML because it will have changed too. The change is due to the new instances we added. Thanks

NikolaiKovalenko:
Could you please send a log, since we do not have access to your project and we cannot see what XML needs to be added
Thanks

danielbeaudreau:
I only have the first failure, for the remaining failures we will need to re-run the integration test.
Ok I'll send it to you via email

NikolaiKovalenko:
file integration_test/data/findscu-instance-expected.xml updated

NikolaiKovalenko:
added newline into findscu-instance-expected.xml file

NikolaiKovalenko:
timeout for DicomWebClientJetty updated

danielbeaudreau:
The timeout changed as we merged PR#110, please rebase this change into the branch

NikolaiKovalenko:
xml files updated with value from logs

danielbeaudreau:
The XML will need to be updated one last time.
I believe the current XML that we have in this PR does not contain all instances, due to the flakyness that previously existed due to missing WaitFor steps.

NikolaiKovalenko:
values in .xml files updated

danielbeaudreau:
FYI I think you can re-run the presubmits by typing a comment "/gcbrun"

danielbeaudreau:
it passed!

danielbeaudreau:
Going to run a couple times to look for more flakes. If they're unrelated to this PR, I won't block the submission
/gcbrun

NikolaiKovalenko:
this is just a miracle

danielbeaudreau:
We're getting flakes resolving plugins from maven - that's not related to the changes.
/gcbrun

danielbeaudreau (changes requested):

PR #101, opened 2020-08-06
Update config for marketplace click-to-deploy solution

George-Wu:
Has following updates:
1. Add examples to some of the parameter fields.
2. Update the version to 0.2.14
3. Remove 'enable exporter' section and add a deploy option that let user choose one from 'importer', 'exporter' and 'both'
4. Add a custom flag to import and export adapter.
5. Add details (e.g. related link) to the descriptions of some fields.

### DIFF
--- a/marketplace_confg/chart/dicom-adapter/templates/application.yaml
+++ b/marketplace_confg/chart/dicom-adapter/templates/application.yaml
@@ -6,7 +6,9 @@ metadata:
   labels:
     app.kubernetes.io/name: "{{ .Release.Name }}"
   annotations:
-    marketplace.cloud.google.com/deploy-info: '{"partner_id": "", "product_id": "", "partner_name": ""}'
+    kubernetes-engine.cloud.google.com/icon: >-
+      data:image/png;base64,{{ .Files.Get "logo.png" | b64enc }}
+    marketplace.cloud.google.com/deploy-info: '{"partner_id": "hcls-dicom-adapter-ctd-public", "product_id": "dicom-adapter", "partner_name": "hcls-dicom-adapter-ctd-public"}'
 spec:
   selector:
     matchLabels:
@@ -18,8 +20,8 @@ spec:
       kind: StatefulSet
   addOwnerRef: true
   descriptor:
-    type: "dicom adapter"
-    version: "0.2.1"
+    type: dicom adapter
+    version: "0.2.7"
     keywords:
       - "dicom"
       - "adapter"
--- a/marketplace_confg/chart/dicom-adapter/templates/application.yaml
+++ b/marketplace_confg/chart/dicom-adapter/templates/application.yaml
@@ -6,8 +6,6 @@ metadata:
   labels:
     app.kubernetes.io/name: "{{ .Release.Name }}"
   annotations:
-    kubernetes-engine.cloud.google.com/icon: >-
-      data:image/png;base64,{{ .Files.Get "logo.png" | b64enc }}
     marketplace.cloud.google.com/deploy-info: '{"partner_id": "", "product_id": "", "partner_name": ""}'
 spec:
   selector:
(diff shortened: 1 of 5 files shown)


marketplace_confg/chart/dicom-adapter/templates/application.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
   labels:
     app.kubernetes.io/name: "{{ .Release.Name }}"
   annotations:
+    kubernetes-engine.cloud.google.com/icon: >-
+      data:image/png;base64,{{ .Files.Get "logo.png" | b64enc }}
```
danielbeaudreau, line 10: does logo.png need to be included in this pr?

marketplace_confg/chart/dicom-adapter/values.yaml
```diff
@@ -2,6 +2,7 @@ import:
   image:
     repo: null
     tag: null
+  deploymeny_type: null
```
danielbeaudreau, line 5: typo here

marketplace_confg/schema.yaml
```diff
@@ -27,129 +27,162 @@ properties:
...
+      title: "importer - Monitoring project id (Optional)"
+      description: >-
+        (Optional) Stackdriver monitoring project id, must be the same as the 
+        project id in which the adapter is running.
+  import.custom_flags:
+      type: string
+      title: "Extra import adapter flag (Optional)"
+      description: >-
+        (Optional) Addtional flag for import adapter.
+        e.g. "--flag_a=valueA"
```
danielbeaudreau, line 121: can we add an example for multiple flags, to clarify if they have to be comma or space separated?

marketplace_confg/schema.yaml
```diff
@@ -27,129 +27,162 @@ properties:
...
-      "(Optional) Stackdriver monitoring project id, must be the same as the project 
-      id in which the adapter is running."
+      (Optional) Stackdriver monitoring project id, must be the same as the project 
+      id in which the adapter is running.
+  export.custom_flags:
+      type: string
+      title: "Extra export adapter flag (Optional)"
+      description: >-
+        (Optional) Addtional flag for export adapter.
+        e.g. "--flag_a=valueA"
```
danielbeaudreau, line 184: here too

marketplace_confg/schema.yaml
```diff
@@ -27,129 +27,162 @@ properties:
...
+    title: "exporter - Pubsub subscription id"
     description: The pubsub subscription id
   export.dicomweb_addr:
     type: string
-    title: "exporter - dicom web address"
-    description: Address for DicomWeb service that sources the DICOM.
+    title: "exporter - Dicom web address"
+    description: >-
+      Address for DicomWeb service that sources the DICOM. e.g. 
+      https://healthcare.googleapis.com/v1/projects/myproject/locations/
```
danielbeaudreau, line 149: It should actually just be this:
"https://healthcare.googleapis.com/v1"
not a full path.

marketplace_confg/chart/dicom-adapter/values.yaml
George-Wu: Changed and tested deployment

marketplace_confg/chart/dicom-adapter/templates/application.yaml
George-Wu: Yeah it's useless here, removed

marketplace_confg/schema.yaml
George-Wu: Unfortunately the marketplace does not support list as input. I tried to input a list in the testing but it failed because it was recognize as a list rather than a string (even the list was in a string format)
In addition, Helm chart does not support string split or read string line by line. If the adapter accepts json string as argument I can add some examples here.

marketplace_confg/schema.yaml
George-Wu: Changed

marketplace_confg/schema.yaml
danielbeaudreau: Got it, ok then it's fine as-is.

PR #114, opened 2020-10-11
Dimse msg lose on down scale statistics added;

oleksii-opanasenko:
STOPPED, CSTORE_REQUEST_PROCESSING_DELTA monitor event added; tests updated;
simple shutdown hook with delta = 0 check added and STOPPED++ metric added;
 DicomServiceException cause Ex changed to ex msg - fix of truncating to  64 symbols;


oleksii-opanasenko:
no need anymore

PR #117, opened 2020-12-05
[Marketplace] Updated marketplace config

DanKotowski:
Updated marketplace config to fix various spelling mistakes as well as update DICOM adapter version.


brenpolley:
Changes to the header texts look good. There are still a few inconsistencies around capitalization in descriptions. We should use "DICOM" and "DICOMweb" unless referring to method names (e.g. dicomWeb.searchForStudies)

DanKotowski:
brenpolley@ I have addressed the casing issues, let me know if you notice anything else that may have been missed.

marketplace_confg/schema.yaml
```diff
@@ -57,131 +57,135 @@ properties:
...
+        (Optional) Addtional flags for the export adapter.
         e.g. "--flag_a=valueA"
 
 required:
 - name
-- namespace
\ No newline at end of file
+- namespace
+- import.dimse_aet
+- export.peer_dimse_aet
```
danielbeaudreau, line 191: We don't know if the user will want to deploy an import or export adapter, so I don't think we should have these as required. Especially not the export peer dimse aet since very few customers use the export adapter.

danielbeaudreau (changes requested):

marketplace_confg/schema.yaml
DanKotowski: I made these required because I don't believe the deployer allows us to deploy each individually, I think both import and export images are  deployed at the same time.  If these are not set, when deploying an error is thrown. Let me know if you still want me to mark these as required?

danielbeaudreau:
I think there was a comment here but it was deleted?

marketplace_confg/schema.yaml
DanKotowski: Removed this requirement.

danielbeaudreau (approved):

PR #123, opened 2020-12-16
Close HTTP response

danielbeaudreau:


jasonklotzer (approved):
Could a try-with-resources be used in these cases?

poojavenkatram (approved):
Thanks Dan for the change and just for documentation on this PR on why this is a good practice: https://stackoverflow.com/a/26541624. 

Best
Pooja

PR #140, opened 2022-02-28
fixes #138

### LINKED ISSUE #138: STOW-RS triplet overwrite capability
Currently, when a C-STORE (DIMSE) request is converted to a STOW-RS (HTTP) request, it will fail with an [HTTP 409](https://cloud.google.com/apis/design/errors#handling_errors) error code if the UID triplet (studyUID/seriesUID/instanceUID) already exits in the DICOM store.

This deviates from the expectation for several PACS systems, which would expect the destination system to accept incoming instances and replace those stored.

By adding this configuration option, the adapter can be configured to internally perform a HTTP POST -> (HTTP 409) -> HTTP DELETE -> HTTP POST, which will effectively delete and replace the instance with the new instance that is being ingested.

jasonklotzer:
/gcbrun

### DIFF
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
@@ -15,27 +15,28 @@
 package com.google.cloud.healthcare;
 
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpMediaType;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
-import com.google.api.client.http.HttpMediaType;
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.api.client.http.InputStreamContent;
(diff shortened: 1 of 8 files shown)


dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClient.java
```diff
@@ -128,18 +147,73 @@ public void stowRs(InputStream in) throws IDicomWebClient.DicomWebException {
...
+   * Reads the resource to figure out the study/series/instance UID path, deletes it and returns
+   * void.
+   *
+   * @param stream The unread stream containing the DICOM instance.
+   */
+  @Override
+  public void delete(InputStream stream) throws IDicomWebClient.DicomWebException {
+    try {
+      // TODO: HTTP 409 content has RetrieveURL as StudyUID, so we need to construct the instance
+      // path. Can we receive the RetrieveURL as the full instance path, as we're constructing here?
```
danielbeaudreau, line 205: Sadly this is not possible, because as you mentioned we only populate Study UID in the URL.

danielbeaudreau (changes requested):
Overall LGTM, one comment

danielbeaudreau (approved):

PR #142, opened 2022-05-01
fixes #141

### LINKED ISSUE #141: STOW retry does not work with multiple DICOM destinations
The recent implementation of STOW retry (#138) does not work with a multi-DICOM destination configuration:
https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/wiki/Routing-to-multiple-DICOM-stores

jasonklotzer:
/gcbrun

### DIFF
--- a/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
+++ b/dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
@@ -9,6 +9,9 @@ import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import org.dcm4che3.data.Attributes;
+import org.dcm4che3.data.Tag;
+import org.dcm4che3.io.DicomInputStream;
 import org.dcm4che3.net.Status;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.http.HttpHeader;
@@ -28,8 +31,6 @@ import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.FuturePromise;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -39,14 +40,17 @@ public class DicomWebClientJetty implements IDicomWebClient {
 
   private static final int CONNECT_PORT = 443;
 
+  private final Boolean useStowOverwrite;
   private final String stowPath;
   private final OAuth2Credentials credentials;
 
   public DicomWebClientJetty(
       OAuth2Credentials credentials,
-      String stowPath) {
+      String stowPath,
+      Boolean useStowOverwrite) {
     this.credentials = credentials;
     this.stowPath = StringUtil.trim(stowPath);
+    this.useStowOverwrite = useStowOverwrite;
   }
 
   @Override
@@ -56,16 +60,6 @@ public class DicomWebClientJetty implements IDicomWebClient {
 
   @Override
   public JSONArray qidoRs(String path) throws DicomWebException {
-    throw new UnsupportedOperationException("Not Implemented, use DicomWebClient");
-  }
-
-  @Override
-  public void delete(String path) throws DicomWebException {
-    throw new UnsupportedOperationException("Not Implemented, use DicomWebClient");
-  }
-
-  @Override
-  public void delete(InputStream stream) throws DicomWebException {
     throw new UnsupportedOperationException("Not Implemented, use DicomWebClient");
   }
 
@@ -155,7 +149,8 @@ public class DicomWebClientJetty implements IDicomWebClient {
       int httpStatus = responseCodeFuture.get();
       if (httpStatus != HttpStatus.OK_200) {
         String resp = resultBuilder.toString();
-        throw new DicomWebException("Http_" + httpStatus + ": " + resp, httpStatus, Status.ProcessingFailure);
+        throw new DicomWebException(
+            "Http_" + httpStatus + ": " + resp, httpStatus, Status.ProcessingFailure);
       }
     } catch (Exception e) {
       if (e instanceof DicomWebException) {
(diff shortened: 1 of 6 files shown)


dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -165,9 +160,125 @@ public void onData(Stream stream, DataFrame frame, Callback callback) {
...
+    return this.useStowOverwrite;
+  }
+
+  /**
+   * Deletes the resource and returns void.
+   *
+   * @param path The resource URL path to delete.
+   */
+  @Override
+  public void delete(String path) throws IDicomWebClient.DicomWebException {
```
jasonklotzer, line 175: Added support for delete inside of HTTP/2 code, primarily for STOW overwrite.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -121,10 +121,16 @@ public static void main(String[] args) throws IOException, GeneralSecurityExcept
...
     IDestinationClientFactory destinationClientFactory = configureDestinationClientFactory(
-        defaultCstoreDicomWebClient, credentials, flags, backupUploadService != null);
+      defaultCstoreDicomWebClient, credentials, flags, backupUploadService != null);
 
-    MultipleDestinationUploadService multipleDestinationSendService = configureMultipleDestinationUploadService(
-        flags, cstoreSubAet, backupUploadService);
+    // NOTE: This is a dependency for cStoreService if stow retry is used or if multiple
+    // destinations are leveraged, due to the backupUploadService dependency.
+    MultipleDestinationUploadService multipleDestinationSendService = null;
+    if (flags.sendToAllMatchingDestinations) {
```
jasonklotzer, line 129: Require explicitly setting this flag for `useStowOverwrite` to work properly, as it relied on the backupUploadService.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
```diff
@@ -165,9 +160,125 @@ public void onData(Stream stream, DataFrame frame, Callback callback) {
...
+  public void delete(InputStream stream) throws IDicomWebClient.DicomWebException {
+    try {
+      DicomInputStream dis = new DicomInputStream(stream);
+      Attributes attrs = dis.readDataset(-1, Tag.PixelData);
+      String instanceUrl =
+          String.format(
+              "%s/series/%s/instances/%s", // NOTE: /studies/ already in stowPath
+              attrs.getString(Tag.StudyInstanceUID, 0),
+              attrs.getString(Tag.SeriesInstanceUID, 0),
+              attrs.getString(Tag.SOPInstanceUID, 0));
```
danielbeaudreau, line 277: Nit: instead of defaulting to a 0 value and potentially deleting something we're not supposed to (instance with IDs of 0), can we throw an exception if these are not found?

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -157,15 +163,12 @@ private static IDicomWebClient configureDefaultDicomWebClient(
       GoogleCredentials credentials,
       Flags flags) {
     IDicomWebClient defaultCstoreDicomWebClient;
+    log.debug("--stow_overwrite set to " + (flags.useStowOverwrite ? "true" : "false"));
```
danielbeaudreau, line 166: nit: instead of having these debug statements in various places, we could have one function that logs the flags state for all of the flags in one place / function.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
danielbeaudreau: Trying to reason about this change a bit: it looks like the state before was, we set the multipleDestinationSendService to null only if the backupService was null. Now we set it to null if sendToAllMatchingDestinations is false. 

I'm wondering, isn't this a breaking change in some ways? E.g. for current users who set up the backupService but do not configure sendToAllMatchingDestinations, won't their backups break now?

danielbeaudreau (changes requested):

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
jasonklotzer: Added issue #143 for this. Removed the debug statement for now.

dicom_util/src/main/java/com/google/cloud/healthcare/DicomWebClientJetty.java
jasonklotzer: Resolved in latest commmit

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
jasonklotzer: Reverted the initialization logic and added a check for multiple destinations + stow retry later in the code.

import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java
```diff
@@ -190,16 +186,22 @@ private static IDestinationClientFactory configureDestinationClientFactory(
...
           multipleDestinations.getLeft(),
           multipleDestinations.getRight(),
           defaultCstoreDicomWebClient);
     } else { // with or without backup usage.
+      if ((flags.destinationConfigPath != null || flags.destinationConfigInline != null)
+          && flags.useStowOverwrite) {
+        throw new IllegalArgumentException(
+            "Must use '--send_to_all_matching_destinations' when using '--stow_overwrite' and"
+                + " providing a destination config.");
+      }
```
danielbeaudreau, line 201: I looked in the cstore service to refresh my memory. So it looks like the condition for this to work is multipleDestinationSendService != null && useStowOverwrite = true, that way we take the path where we use the backup service, and can use the delete and re-store logic. [1] Just wondering why it matters if --send_to_all_matching_destinations is true vs. just using one destination here? 

Can we simplify this check to:
if multipleDestinationSendService == null && useStowOverWrite, then throw error.

[1] https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/c893c51a4cce514e10511fb9d53f9205e60c2f17/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java#L119

jasonklotzer: Two reasons: 
1) The import adapter always [initializes](https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/master/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java#L129) the multipleDestinationSendService, so it won't be `null`.
2) This one took me a bit of time to understand, but this [code](https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/master/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java#L181) initializes the `destinationClientFactory` based on send_to_all_matching_destinations, and then the the `Single/MultipleDestinationClientFactory` will assign the destination to the `DestinationHolder` based on the C-store requests. The problem is the single/multiple implementations actually assign the values based on different properties and the values have to be retrieved via different properties (namely `getSingleDestination()` vs `getHealthcareDestinations()`). And then this [code](https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/master/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/CStoreService.java#L129) relies on the `DestinationHolder` (though `multipleSendService` will never be null in import adapter mode).

danielbeaudreau: For 1. It looks like we do sometimes not initialize multipleDestinationSendService, if backupUploadService == null: 
https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/195a0af70b1fed12df24575fad7b1375faff63d5/import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/ImportAdapter.java#L220

For 2. hmm this is interesting. So it looks like there's a bug in the code? When:
- the single destination client factory is specified
- backup service is initialized
- but sendToAllMatchingDatasets is not specified

then we will use multipleDestinationSendService and call destinationHolder.getHealthcareDestinations() in CStoreService.Java.
However destinationHolder.getHealthcareDestinations() will not return what we expect, since single destination client factory only returns anything on getSingleDestination().

danielbeaudreau: The bug in 2. being that we would just be sending to all destinations rather than the single destination that we intended to send based on the filter. And the filtering in the single destination would not work. 

Am I correct? The code is a bit complex here so I could have got this wrong. Anyway I think that issue is outside of the scope of this PR, but I wanted to get on the same page with this before deciding what to do about it.

jasonklotzer: As discussed, going to leave the code as-is for now. Created issue #144 for 2.

Issue #120, opened 2020-12-14
Feature request: Under a config flag, decouple STOW-RS request from DIMSE C-STORE-RQ handling


jasonklotzer:
Goal: Increase the amount of images/second that can be handled by a single association on a single instance of the adapter.

In load testing of an environment where the adapter is running on-prem in a VM, it was found that a single DICOM association is not able to process more than 2-3 images/second.

The underlying assumption is that this is due to the synchronous nature in which the adapter handles a single store request per association:
Request sequence:
(PACS) -> C-STORE-RQ -> (Adapter) -> STOW-RS -> (Healthcare API)
Response sequence:
(PACS) <- C-STORE-RES <- (Adapter) <- HTTP Code <- (Healthcare API)

This is further supported by the fact that the P50 of the HCAPI endpoint is in the 100ms range, yielding an absolute max throughput per association of 10 images/second, ignoring all other portions of the insertion pipeline.

The suggested approach to improve the amount of image/second/association is to decouple the flow of data between (PACS) and (Adapter) from that of the (Adapter) and (Healthcare API), via thread pool or some other primitive with asynchronous support.

jasonklotzer:
@danielbeaudreau

luxunxiansheng:
Hi, there, 
       Is there any timeline on this enhancement to be done ?    Many thanks

L.B.

jasonklotzer:
Apologies for the delay. This feature is under code review and should be available shortly.

PR #145, opened 2022-07-14
fixes feature request #120

jasonklotzer:


import/src/main/java/com/google/cloud/healthcare/imaging/dicomadapter/cstore/multipledest/MultipleDestinationUploadService.java
```diff
@@ -93,26 +103,64 @@ public void start(ImmutableList<IDicomWebClient> healthcareDestinations,
...
+
+  private void removeLazyResolvers(int associationId) {
+    HashMap<String, List<CompletableFuture>> resolverMap = lazyFutureResolvers.get(associationId);
+    if (resolverMap != null) {
+      log.debug("Cleaning up " + resolverMap.size() + " files for association " + associationId);
+      resolverMap.forEach(
+          (sopInstanceUID, futures) -> {
+            futures.forEach(CompletableFuture::join);
+            backupUploadService.removeBackup(sopInstanceUID);
+          });
```
danielbeaudreau, line 162: This waits for the CompletableFuture to finish before removing the backup, correct?

danielbeaudreau: Also any error reporting we want to do here if the future failed or threw an exception?

jasonklotzer: Yes, the original code waited for all the futures to complete. Unfortunately it silently masked exceptions, so I broke out the code to explicitly handle any exceptions, similar to the synchronous code.

PR #171, opened 2024-05-14
Allow configuration of export client AET.

craigatron:
This allows specifying the presented AET in cases where the peer is expecting something other than `EXPORTADAPTER`.


craigatron:
thanks for the quick review!

PR #174, opened 2024-07-29
Vendor the deid-redactor library

craigatron:
had to make a few changes to the redactor library:

- deleted settings.gradle
- moved repositories config into individual packages
- bumped protobuf plugin to version that supports gradle 6


craigatron:
@IvanIsCoding think this should be good to go now!

IvanIsCoding (approved):

PR #176, opened 2025-01-09
Include Query/Retrieve Level in C-FIND Response

ncolossi:
This PR fixes an issue where the C-FIND response was missing the required Query/Retrieve Level tag when returning study results (status=Pending), causing errors in viewers like Carestream's. The missing tag has been added and is populated with the value from the C-FIND request, as per the DICOM standard.


IvanIsCoding:
The integration test is failing at Step 1 in one of the `check-diff` steps: https://github.com/GoogleCloudPlatform/healthcare-dicom-dicomweb-adapter/blob/32a96401b21eb7d733bff799fb5bbd8553d85f16/integration_test/step_1.py#L68

You need to update the expected output to include the new field added in the response I assume?
```
```

jasonklotzer (approved):
lgtm

